PR #30, opened 2018-07-31
Removes skill check for overwatch consoles

chesse20:


code/modules/cm_marines/overwatch.dm
```diff
@@ -36,7 +36,7 @@
 	if(..())  //Checks for power outages
 		return
 
-	if(user.mind.cm_skills && user.mind.cm_skills.leadership < SKILL_LEAD_EXPERT)
+	if(1 == 2) //i think this disables the check to see if you have the training to use the overwatch console
```
Jared-Fogle: genius

Ziiro: 200 IQ code

ninjanomnom: Just remove the code

chesse20:
@ninjanomnom done

ninjanomnom:
Skill checks have a place atm and just plain removing them without something to fill their place like ids seems a bad move. Skill checks that when removed don't need an id check are fine to be removed but stuff like this I'm not so sure about.

chesse20:
the overwatch consoles are in an area where only command staff have access anyways, and the worst thing that could happen is a marine breaking into command and then yelling woodys got wood over the overwatch channel and/or firing an orbital beacon

Wilchenx:
>the overwatch consoles are in an area where only command staff have access anyways, and the worst thing that could happen is a marine breaking into command and then yelling woodys got wood over the overwatch channel and/or firing an orbital beacon

so why would we want to ever let that happen?

chesse20:
@ninjanomnom alright it works

PR #36, opened 2018-07-31
arcade machine rebalances

chesse20:
changes the numbers to tg station numbers and also i uncommented out the code that gives it a random name


chesse20:
i added the requested line back in

PR #42, opened 2018-07-31
Made ghosts move instantly

Jared-Fogle:
Playing as an observer is annoying in CM because it only updates your position once every 1.5s. With this change, your position will update as the movable atom moves. There's no components in the CM codebase so the code is a bit gross.

### DIFF
--- a/code/_onclick/observer.dm
+++ b/code/_onclick/observer.dm
@@ -26,7 +26,7 @@
 
 		// Otherwise jump
 		else
-			following = null
+			unfollow()
 			loc = get_turf(A)
 
 		return 1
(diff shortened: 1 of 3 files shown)


MrStonedOne:
There are some things that don't call moved(), mainly teleports and other direct loc changes.

The quickest way would be to just make the sleep use 1.

The proper way would be to make a process that loops thru all following ghosts and updates their loc

The proper proper way would be to combine the above with how you are currently doing it, so the process is just a backup.

The /tg/ Proper way would be to force everything to not assign locs and instead call some forceMove proc that ghosts can bind to.

The /tg/ Proper Proper way would be components

Jared-Fogle:
What I'll probably do is keep the old sleep thing as a backup.

Jared-Fogle:
@MrStonedOne Done

PR #45, opened 2018-07-31
Removes all delays in putting on items

Mothblocks:
Delays in putting on armor are stupid.


Wilchenx:
i think it was made to discourage people taking off armor to run away faster from xenos then putting it back on

Wilchenx:
anyway i don't see a big issue in removing this

MrStonedOne:
I don't think removing these flat out is the right answer, but I do agree they might be a bit too long.

PR #53, opened 2018-07-31
Gets rid of automatic spam filter

Suicidalpickles:
punishing people for typing too quickly is pretty dumb; this removes the delay on talking in ic/ghostchat/me. Im also a fool and accidentally left the other commit in this.


epicmemename:
speedmerge

PR #63, opened 2018-08-01
mapmerge2 + converts maps + adds mapmerge shit to .gitignore

praisenarsie:
thanks to @spacemaniac for writing mapmerge2

someone should port the .gitignore from tgstation


praisenarsie:
don't understand the issue

SpaceManiac:
Double-check that the maps still open in Dream Maker after being stored as TGM. If they crash they have too many lines in them for BYOND to handle (yes, it's actually based on line count) and will need to be stored non-TGM. It's possible to edit the mapmerge script to output non-TGM files to get a medium between messy maps and too-large maps.

praisenarsie:
looks like the almayo was 2big

Jalleo:
can you also get the other maps in the repo able to be loaded as well? 
They need some path changes. Some small others larger.

(If we get asked to remove em we can do so)

Rohesie:
Is there still interest here?

ninjanomnom:
Aparently not

PR #67, opened 2018-08-03
Reverts the removal of the different license files.

kyrahabattoir:
Given that the AGPL only applies to the collection of work, yet its individual elements retain their original licensing, these shouldn't be removed.

### DIFF
--- a/LICENSE-GPLv3
+++ b/LICENSE-GPLv3
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The GNU General Public License is a free, copyleft license for
+software and other kinds of works.
+
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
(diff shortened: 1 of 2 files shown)


MrStonedOne:
Without anything that states what they apply to they mean nothing.

Things that mean nothing shouldn't be the repo.

kyrahabattoir:
Then you should probably remove the LICENSE file you added too since it doesn't state that it only applies to the Proc Scheduler.

MrStonedOne:
and yet some how /tg/ managed to state what licenses apply to where and when without modifying the license files.

kyrahabattoir:
You're gonna keep doing whatever the fuck you want and making it up as you go until someone actually sue you don't you MSO?

MrStonedOne:
No, I'm going to be obtrusive until you actually attempt to solve the issue I brought up.

You are making a pr, I am the repo maintainer, I have brought up an issue with your pr, you add new license files without stating what they apply to.

Maybe rather than issue adhom attacks to me you can figure out some shit to add to the readme.md to make these files more clear.

Protip: Assuming bad faith can be itself a bad faith act.

MrStonedOne:
This PR will be closed in 72 hours unless the feedback is addressed.

Please add a note to readme.dm explaining what license files apply in what situation.

You may self close this PR if you think you can not address the feedback within the next 72 hours, doing so allows  you to self open the pr, rather than wait for a maintainer to do it.

kyrahabattoir:
I'm not adding anything I'm reverting a honest mistake.

PR #70, opened 2018-08-05
activates trow mode when turning on a flare

### LINKED ISSUE #19: Flares don't automatically set you to throw when used
There is never a situation where you won't use one as a light grenade.

Tlaltecuhtli:
closes #19 
it should work


Wilchenx:
make it a toggle in prefrences

Tlaltecuhtli:
why

ninjanomnom:
I don't see a reason for this. People can activate throw on their own.

PR #92, opened 2018-08-14
Replaces the shitty micro-core with a better designed one

francinum:
It's replaced the fore firing systems control room. not like that actually did anything. This clears some room for future space filler next to tcomms.


Rohesie:
Got pictures of this?

francinum:
[Map Diff check result](https://github.com/ColonialMarines-Mirror/ColonialMarines-2018/pull/92/checks?check_run_id=12944609)

Rohesie:
This doesn't really offer an improvement and actually makes the place look a bit worse. For now it's denied.

PR #96, opened 2018-08-15
Suicide PR 2.0, now without the restrictions and with useful admin notifications!

F-OS:
This PR re-implements suicide with a admin notification if you suicide as a important job.

### DIFF
--- a/code/game/verbs/suicide.dm
+++ b/code/game/verbs/suicide.dm
@@ -24,7 +24,7 @@
 		if(src.mind && src.mind.assigned_role != "Squad Marine") // We don't care about squad marine suicides or people suiciding without minds.
 			message_admins("[ckey] as suicided as a [src.mind.assigned_role]", 1)
 		
-		if(do_after(user, 40, TRUE, 5, BUSY_ICON_HOSTILE))
+		if(do_after(src, 40, TRUE, 5, BUSY_ICON_HOSTILE))
 			suiciding = 1
 			var/obj/item/held_item = get_active_hand()
 			if(held_item)
--- a/code/game/verbs/suicide.dm
+++ b/code/game/verbs/suicide.dm
@@ -16,64 +16,66 @@
 		return
 
 	var/confirm = alert("Are you sure you want to commit suicide?", "Confirm Suicide", "Yes", "No")
-
+	src.visible_message("<span class='warning'>[src] gets a horrible look in their eyes!</span>")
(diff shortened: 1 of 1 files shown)


Wilchenx:
Letting people suicide while getting tacklespammed before infection is super fucking bad.

You can suicide anyway with a pistol to mouth/chest.

We don't need this and this will just break shit.

F-OS:
@Wilchenx You can't suicide while stunned from what i can see.

Rohesie:
I cannot see a single positive thing this adds to the game, and I can see a series of issues it causes:
* Harder for xenos to capture hosts, easier for them not to make it alive into the hive to be infected. Xenos already have the issue of being too slash-happy.
* Kind of much easier for people to die with a command away, blocking certain job slots that unlike in TG cannot really be replaced due to the skills they carry.
* A definite descent into the lower RP spectrum.

I love TG and have been playing it for half a decade. It has many good features CM could use. This is not one of them. It offers nothing besides "muh TG culture".

Wilchenx:
@F-OS yeah but what if you break out of a stun for a second to pull this off


And why do we need this anyway if there's cryo and you can still suicide without a lazy verb.

F-OS:
Based on some discord feedback I've added a delay to suicide. However, I do feel that this should be added because of some reasons I've elaborated on in discord.

Wilchenx:
image

If thats the only reason for this then you can just close it and save some work for ninja

F-OS:
@Wilchenx I have some other reasons, First of all, arbitrary restrictions on stuff like this is annoying and unneeded, along with that, you can already commit suicide in a multitude of ways so keeping this one way locked up behind a antagonist check makes absolutely no sense, and what you referenced there isn't even relevant as i realized the point Rohesie was making and added a time delay.

Rohesie:
As far as I know it was intended to be higher RP than TG.
This is not an arbitrary restriction. It is a lack of an arbitrary feature. You _can_ suicide with a gun to the mouth (which is designed thinking of captured host/xeno interaction balance) and you can use other means of causing a lot of damage for yourself. And xenos have ways to counter that (the resin nest being a health stabilizer).
There is zero reason to add the feature of holding your breath to death, which as funny as that is in a TG context, is simply out of place here.

F-OS:
This is overwhelmingly hated so until i get a opinion on this from rustled/mso that's a clear yes/no i'm closing this.

PR #101, opened 2018-08-16
Makes the suicide rejection message friendlier

JamesDenholm:
Mechanically, changes nothing about the suicide verb.

BUT, my gut feeling is that a lot of the desire to change the mechanic stems from... powergaming nerds.

BUT some of it probably stems from the message making it sound like suicide isn't permitted ever-ever-ever and will get you the bwoink for goodness, and there _are_ plenty of situations in your average round where suicide is a RP-reasonable course of action -- it's not even remotely difficult to imagine a character who would off themselves after breaking out of a nest, for example, and that sort of thing probably shouldn't be discouraged.

### DIFF
--- a/code/game/verbs/suicide.dm
+++ b/code/game/verbs/suicide.dm
@@ -21,7 +21,8 @@
 
 	if(!permitted)
 		message_admins("[ckey] has tried to suicide, but they were not permitted due to not being antagonist as human.", 1)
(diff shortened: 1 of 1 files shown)


F-OS:
Cut out the admin message too, there is no spam check on suicide from what i can see and there's no reason to notify admins about failed suicides, it only serves to annoy.

JamesDenholm:
It's there so admins can be aware someone made the attempt. A useful "this person might be a prat" signal.

PR #105, opened 2018-08-17
Makes everything tg: Hotkeys & buttons edition

LetterN:
after this: makes all normal lameui to greyui (the typical tg ui (not tgui))
now you can ""accidentally"" ick in ocky when using hotkeys

### DIFF
--- a/interface/interface.dm
+++ b/interface/interface.dm
@@ -67,7 +67,9 @@ Hotkey-Mode: (hotkey-mode must be on)
 \tq = drop
 \te = equip
 \tr = throw
+\tm = me
 \tt = say
+\to = ooc
 \tx = swap-hand
 \tz = activate held object (or y)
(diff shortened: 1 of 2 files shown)


MrStonedOne:
Could you re-add the thing to make the text bar white when hotkey mode is enabled. it was a really helpful visual cue.

subject217:
Also, no resist hotkey? It's bound to B on /tg/.

Rohesie:
Once the author addresses these requests, either denying or accepting them and making the changes, I'm merging this. It has been tested and works as intended.

subject217:
If LetterN doesn't want to do it I'll open a PR with that and probably some other binds too.

Rohesie:
The PR has been tested and approved by the players through in-game vote. Further modifications can be made in a new PR. This is accepted.

PR #106, opened 2018-08-17
Reallows slipping 

F-OS:
Slipping was originally disabled on CM due to roundstart shenanigans that didn't fit their targeted roleplay environment, due to the lower RP this server is shooting for i feel slipping should be allowed, and if people get griefy with it then that can be handled by the admins.

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -1418,7 +1418,7 @@
 	T.add_vomit_floor(src)
 
 /mob/living/carbon/human/slip(slip_source_name, stun_level, weaken_level, run_only, override_noslip, slide_steps)
-	if(shoes && !override_noslip) && (shoes.flags_inventory & NOSLIPPING)) // Slipping is saved!
+	if((shoes && !override_noslip) && (shoes.flags_inventory & NOSLIPPING)) // Slipping is saved!
 		return FALSE
 	. = ..()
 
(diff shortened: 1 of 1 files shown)


F-OS:
am i retarded?

LetterN:
slippery slope

Rohesie:
This is HvX, not HvH. Slipping still works if you have no boots. Literally why? This change offers no feature _besides_ griefing.
It's not like water is readily available like in TG, or as if it would fit the MRP theme.

epicmemename:
considering "human versus human" interactions aren't immediately bannable like it was where you came from, slipping serves an actual purpose for area denial / traps / comedy. @Rohesie

Rohesie:
I mean, yes, but it's just another way to antagonize the team. It's as if you added a PR making the stun talisman also be able to stun other blood cultists. I mean, why? The objective here is to defeat the opposing team, not to in-fight.

epicmemename:
marines aren't a team in some cases, and human vs human conflict is bound to happen. Instead of removing complexity from potential scenarios, you could make it tailored towards where the scenarios would normally happen

eg: you can only make floortiles slippery
it's much more obvious 

+ marines are technically not a team, and cultists killing other cultists who are being antagonistic does happen

Sifaus:
Please no

LetterN:
incoming drama

octareenroon91:
👌

Rohesie:
If you want to port /tg/'s slippery system and add `clothing_flags |= NOSLIP` (or a different method) to marine boots instead, be my guest. That would bring us closer to /tg/ while allowing us to retain some differences.
But the current change would just make the admin's problems higher without adding anything substantial to the game.
Slipping is fundamental to /tg/ and I could not see that game without it. But it has no place in current DMCA.
Unless I'm overruled by @ninjanomnom , I'm closing this PR.

PR #117, opened 2018-08-17
Revert Lurker's name back to Hunter

avosirenfal:
This caste was named Hunter for like three years and they changed it a month before the source leaked. UNDO. GROSS.


Rohesie:
I actually prefer Lurker. It describes the caste better. Specially since the real stealth changes, which could be improved.

ninjanomnom:
I see no problem with lurker and this is just code churn

PR #123, opened 2018-08-18
Allow benos to attack each other after the round ends

avosirenfal:
End of round grief.


Rohesie:
Add an attack message as well.

PR #124, opened 2018-08-18
Remove build check

MrStonedOne:
This was a silly idea. let us never speak of it again.

### DIFF
--- a/code/modules/mob/new_player/new_player.dm
+++ b/code/modules/mob/new_player/new_player.dm
@@ -19,7 +19,7 @@
 		mob_list += src
 
 	proc/version_check()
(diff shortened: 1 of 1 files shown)


JamesDenholm:
Are we sure the supposed exploit doesn't actually exist, tho?

ninjanomnom:
There are occasionally exploits but it's preferable to block individual builds instead of this

MrStonedOne:
There's still an exploit. this check is in the lobby. anybody can join with a good client get past the lobby, then disconnect and reconnect with one of the broken clients, you'll skip the lobby so it will never check the build.

JamesDenholm:
Oh shit, wow.

MrStonedOne:
I'd fix it but that requires editing multiple files

PR #126, opened 2018-08-18
Add a red bandana

sabiram:
Brother


Jared-Fogle:
Why do we still not have the icon diff bot?

PR #128, opened 2018-08-18
ooc verbs

LetterN:
and removes that anti-spam


Jared-Fogle:
You removed the anti-spam say stuff but didn't document it

PR #133, opened 2018-08-18
Toughens some of the windows on the main deck

francinum:
Anti-greytide at best.


F-OS:
Can you replace the snowflake requisitions windows with this please?

PR #137, opened 2018-08-18
Refactor __DEFINES, __HELPERS, _onclick for conformity to style guide and legibility

bobbahbrown:
Removed unnecessary whitespace, relocated some nested comments to be the preamble for the relevant methods, refactored absolute paths.

This is the current version of my regex for refactoring for style:
`(^([\t ]*(datum\/|proc\/|verb\/|client\/))|(^(\n){2,})|(^((datum|proc|verb|client)\s))|(\/.*\)[ ]*((\/\/).*)*(\n\s\n\s)+)|^([\t ]+(?!for|if|var)[^\(\s\.]*\([^\(\s0-9]*\/[\S ]*\)))`

edit: please note I'm currently not replacing by regex, only finding by regex.

### DIFF
--- a/code/__DEFINES/__game.dm
+++ b/code/__DEFINES/__game.dm
@@ -74,7 +74,6 @@ These are used with cdel (clean delete). For example, cdel(atom, TA_REVIVE_ME) w
 //Object specific defines
 #define CANDLE_LUM 3 //For how bright candles are
 
-
 //Preference toggles//
 //toggles_sound
 #define SOUND_ADMINHELP	1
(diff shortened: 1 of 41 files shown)


code/__DEFINES/__game.dm
```diff
@@ -74,7 +74,6 @@ These are used with cdel (clean delete). For example, cdel(atom, TA_REVIVE_ME) w
 //Object specific defines
 #define CANDLE_LUM 3 //For how bright candles are
 
-
```
ninjanomnom: I see no reason to make pure whitespace changes like this

code/_onclick/hud/robot.dm
```diff
@@ -2,7 +2,7 @@
 	..()
 	var/obj/screen/using
 
-//Radio
+	//Radio
```
ninjanomnom: These sorts of formatting changes are also unnecessary, if someone wants to modify it when coding that's fine but we don't have style guidelines for where the comment needs to go.

PR #143, opened 2018-08-18
Possibly fixes #119

### LINKED ISSUE #119: Vents aren't visible to aliens ventcrawling
This is very painful for aliens. Example image: There's a straight line of vents here that's unbroken, I'm crawling in them but they're not visible.

image

kachnov:
When ayys are ventcrawling vents will appear above the pipes now, I have no clue if this actually works but if its a layering issue (newer versions of BYOND have a few) it might do something

closes #119

### DIFF
--- a/code/_onclick/ventcrawl.dm
+++ b/code/_onclick/ventcrawl.dm
@@ -126,11 +126,14 @@ var/list/ventcrawl_machinery = list(/obj/machinery/atmospherics/unary/vent_pump,
 	if(!network)
 		return
 	for(var/datum/pipeline/pipeline in network.line_members)
-		for(var/atom/X in (pipeline.members || pipeline.edges))
+		for(var/X in (pipeline.members || pipeline.edges))
 			var/obj/machinery/atmospherics/A = X
 			if(!A.pipe_vision_img)
 				A.pipe_vision_img = image(A, A.loc, layer = BELOW_MOB_LAYER, dir = A.dir)
 				A.pipe_vision_img.alpha = 180
+			for (var/obj/machinery/atmospherics/unary/vent_pump/V in get_turf(A))
+				A.pipe_vision_img.layer = V.layer - 0.01
+				break
 			pipes_shown += A.pipe_vision_img
 			if(client)
 				client.images += A.pipe_vision_img
--- a/code/_onclick/ventcrawl.dm
+++ b/code/_onclick/ventcrawl.dm
@@ -126,14 +126,11 @@ var/list/ventcrawl_machinery = list(/obj/machinery/atmospherics/unary/vent_pump,
 	if(!network)
 		return
 	for(var/datum/pipeline/pipeline in network.line_members)
-		for(var/X in (pipeline.members || pipeline.edges))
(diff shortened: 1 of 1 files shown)


code/_onclick/ventcrawl.dm
```diff
@@ -126,11 +126,14 @@ var/list/ventcrawl_machinery = list(/obj/machinery/atmospherics/unary/vent_pump,
 	if(!network)
 		return
 	for(var/datum/pipeline/pipeline in network.line_members)
-		for(var/atom/X in (pipeline.members || pipeline.edges))
+		for(var/X in (pipeline.members || pipeline.edges))
```
ninjanomnom: Do these lists only have atoms for sure? I wouldn't be surprised if there are nested pipeline datums

kachnov: those lists are /obj/machinery/atmospherics/pipe so unless someone fucked up, yes

code/_onclick/ventcrawl.dm
```diff
@@ -126,11 +126,14 @@ var/list/ventcrawl_machinery = list(/obj/machinery/atmospherics/unary/vent_pump,
...
 			var/obj/machinery/atmospherics/A = X
 			if(!A.pipe_vision_img)
 				A.pipe_vision_img = image(A, A.loc, layer = BELOW_MOB_LAYER, dir = A.dir)
 				A.pipe_vision_img.alpha = 180
+			for (var/obj/machinery/atmospherics/unary/vent_pump/V in get_turf(A))
+				A.pipe_vision_img.layer = V.layer - 0.01
+				break
```
ninjanomnom: This is frankly a terrible way to fix the problem, the vents should just be a slightly higher layer in the first place

kachnov: The vent layer is about 25 defines lower than BELOW_MOB_LAYER (2.53 vs 3.79), changing it to be above BELOW_MOB_LAYER could introduce a lot of inconsistencies. And I think the image being at BELOW_MOB_LAYER is generally nice (the vent images go above doors), so it's also a bad idea to give it a lower layer than vents. This is the solution that will fuck up as little as possible IMO.

ninjanomnom: The issue isn't so much with the layering as it is you're looping through every vent on every turf with an atmospherics machinery. Making the pipe images be at their proper layer from the get go without having to be modified every time per mob is the only proper way to handle it.

kachnov: Its probably better to close this PR if the only way to fix it is changing the pipe layer to a lower one. IIRC even on /tg/ ventcrawling pipe images appear above doors and stuff, changing this will remove that helpful feature.

There's probably a better way to fix the issue judging by [this](https://gitlab.com/cmdevs/ColonialMarines/issues/3654) report

ninjanomnom: You can also change it to a higher one, the exact layer isnt what matters

kachnov:
closing because this is a poor solution to the issue

PR #85, opened 2018-08-10
Skills

Rohesie:
* Remove the majority of the skill locks, substituting them for ID locks , fumbling time or other penalties.
* Tank access made. Tank is ID locked.
* Dropship had spaghetti locks. Now they are only locked by ID, while lack of pilot skill forbids from optimizing the route and makes operating it slower.
* Riot grenade launchers are ID locked to MPs and Spec weapons to specs. Those with a mind skill can use them without an ID, for Wey-Yu PMCs and other non-marine humans' access is unfinished.
* Smartguns have no locks, but the non-trained user will suffer heavy recoil and accuracy penalties.
* Scouts can use shotguns again.

PR #146, opened 2018-08-19
Big Weapons Rebalance

Rohesie:
Done:
* Pointblank fixed. It had a bug in where the queen/crusher crest was bypassed due to the bullet not having a location nor traveled distance to calculate falloff damage. The +10% damage reduced to weaken a tiny bit the buckshot meta. A melee attack before firing has been added so the bayonet is more useful. Burstfiring pointblank bug is gone, you can stab and spray now.
* Buckshot still kicking ass. Can one-shot in pointblank half the xenos but it's a little less OP than before. A little less falloff to make it useful in mid-range, but a little more negative AP to be less useful against highly-armored targets.
* Flechettes are useful now. Buff in damage and AP. Solid against armored targets, though they spread.
* SMGs buffed in power, but have higher falloff. That means they are powerful at close range, but can't compete with a rifle in the long one.
* Revolver buffed. Main ammo remains as inaccurate and weak as ever, but it has a special effect: stagger and slow. The target won't be able to use special abilities (T1s and T2s only) and will get a tiny slowdown. Both for a small time, but it stacks. It's technically possible to outduel a xeno with akimbo revolvers, but you'll gave to git really good to do that. Marksman ammo had negative AP. Negative AP. Fixed that outrageous monstrosity and turned it positive, as it was probably intended and its description suggested. It's for long (scope) range.
* Slugs don't get a damage boost but they do get the stagger and slowdown effects added. More tactical than damage-dealer.
* Pistol damage and accuracy buffed. They offer the sidearm alternative for medium range, but still a sidearm. Don't expect it to really compete with anything serious. Hollowpoint is similar to a revolver bullet in effect. The rare highpower automag has the shooting speed and stunning effect of a slug-loaded tactical shotgun, but much less accuracy and damage.
* PPSh fires like an inaccurate and weak machinegun from hell. It's fun even if not horribly effective. Though you can kind of kill weak T1s with it.
* The tactical shotgun has been nerfed a little. Less accuracy and damage for the firing speed, not to make it OP with the staggerstun. The HG shotgun survivors get has gotten extra accuracy and lowered scatter at the cost of a little firing speed. As they can get attached a miniscope they are more useful for defenses, as it fits the survivors.
* Extra projectiles now have a scatter chance. Buckshot is still 100% as before, but flechettes are 80% until I make the system a little more elegant.
* Removed two errors which were preventing attachments from affecting falloff damage.
* Rifle regular ammo has been given a little AP to make it competitive and simulate the higher power rifles offer in comparison with other bullets. Not a huge change in statistics. Rifle AP has had its damage lowered so it's less competitive against lowly armored targets than regular ammo, but its penetration increased so it can serve as a queen and crusher counter now that pointblank buckshot is gone. Still won't get you far if you keep shooting their crests though. Rifles get falloff halved, so their long-range is a little more viable.
* Wield delay turned into a soft one: you wield and have a unwield stats until the delay ends, not simply fail to wield. Busy icon added while the marine is stabilizing the wielded gun so you know when it's safe to shoot accurately.
* Attachments reworked: suppressor falloff reduced; bayonet increases size of the gun a little (due to pointblank buff); extended barrel as well, but also reduces scatter a little; recoil compensator has a smaller damage reduction; quickfire shooting speed reduction halved to make it again an option; magnetic harness no longer reduces accuracy; scope now severely increases the burst delay, not just any kind of shooting speed, also causes wield delay; mini-scope has a smaller burst delay; stocks no longer lower the fire rate, but increase the wield delay instead; bipod greatly reduces burst spread and no longer slows down the rate of fire at the cost of a considerable wield delay and increased size of weapon.
* Bipods made more intuitive. They deployed anywhere and gave you penalties but no benefits unless when near a valid support, which you had no feedback about. It now only deploys near a proper support and if you fire it away from one it will retract, warning you and avoiding you further unwanted penalties.

To do:
* Bipod retracts once the user moves, unwields or drops the weapon. Animation for the deployed bipod.


Borisvanmemes:
Doing god's work eh?
also the hell is a PPSh?

Ledwin1233:
Fix the checks big boy

F-OS:
negative ap
>negative ap
**_### negative ap_**

Holy fuck CM coders how do you sleep at night

Lokiki01:
**Scout buff**

- Able to use shotguns again
- Add a bit more firestacks to incendiary ammo
- Return BFA attachment to rifle(It was one of the first nerfs, might be a bit too strong with 4 round burst)
-Add ability to load two magazines into rifle. Swapping between them is unique-action

Lokiki01:
**Sniper buff**

-A bit better ROF (like 0.5 fire delay)
-Return scope usage on load-from-attachment verb
-Add firestacks on incendiary bullets
-Add a bit of damage to marksman ammo, so spec would be able to two-shot mature runners at least
-Rework flak to 'explode' on the tile you click, dealing damage in 8 tiles around. Damage is the same as from marksman bullet but with much less AP

Lokiki01:
**RPG buff**

-Remove explosions from AT rockets, and add damage so it will be able to crit high tier xenos
-Remove 1 tile radius from HE rockets, add more stun and more damage. 
-WP rockets, should leave a bigger flame field after explosion, with strength of flames compared to blueflames
-With buffed AT tockets mini-scope should be removed probably

subject217:
pull requests aren't the right place to suggest balance changes, if you had minor ideas on how to adjust rohesie's existing changes it'd be okay but pretty much everything you've said is way out of scope

Rohesie:
Spec weapons will get their own update eventually as they are still viable and there are more urgent changes to be done. There is a lot to fix, but those suggestions are taken into account for the future.

Scout shotgun-use restriction has been removed here: https://github.com/ColonialMarines-Mirror/ColonialMarines-2018/pull/85
It was a skill-based restriction and thus belongs in that update.

Rohesie:
SMG fires really fast already. You should try it in single-fire to see. If not, the 4-round burst isn't bad either.

ColonialMarinesALPHA.dme
```diff
@@ -1465,6 +1465,5 @@
 #include "interface\skin.dmf"
 #include "maps\Z.01.LV624.dmm"
 #include "maps\Z.02.Admin_Level.dmm"
-#include "maps\Z.03.USS_Almayer.dmm"
```
F-OS: what

Rohesie: I swear to god, what the heck?

code/modules/projectiles/updated_projectiles/gun_helpers.dm
```diff
@@ -191,6 +191,46 @@ As sniper rifles have both and weapon mods can change them as well. ..() deals w
...
 
+/obj/item/weapon/gun/proc/wielded_stable() //soft wield-delay
+	if(world.time > wield_time)
+		return 1
+	else
+		return 0
+		
+/obj/item/weapon/gun/proc/do_wield(mob/user, wield_time) //a poor way to make timed actions show an icon without affecting them until /tg/'s do_mob is ported.
+	var/delay = wield_time - world.time
+	if(!istype(user) || delay <= 0) return FALSE
```
ninjanomnom: Single line control statements like this are a no

code/modules/projectiles/updated_projectiles/gun_helpers.dm
```diff
@@ -191,6 +191,46 @@ As sniper rifles have both and weapon mods can change them as well. ..() deals w
...
+	var/mob/living/L
+	if(istype(user, /mob/living)) L = user
```
ninjanomnom: same as above

code/modules/projectiles/updated_projectiles/gun_system.dm
```diff
@@ -702,34 +707,43 @@ and you're good to go.
...
+								BP.distance_travelled = get_dist(user, M)
+								BP.ammo.on_hit_mob(M, BP)
+								M.bullet_act(BP)
+								cdel(BP)
+
+						projectile_to_fire.ammo.on_hit_mob(M, projectile_to_fire)
+						M.bullet_act(projectile_to_fire)
+						last_fired = world.time
+
+						if(!delete_bullet(projectile_to_fire)) cdel(projectile_to_fire)
```
ninjanomnom: ^

code/modules/projectiles/updated_projectiles/gun_system.dm
```diff
@@ -601,9 +602,13 @@ and you're good to go.
 				projectile_to_fire.accuracy *= config.base_hit_accuracy_mult + config.hmed_hit_accuracy_mult //More accuracy.
 				recoil_comp-- //Less recoil.
 				scatter_chance_mod -= config.med_scatter_value
-				burst_scatter_chance_mod = -2
+				burst_scatter_chance_mod = -3
 				if(prob(30)) projectile_to_fire.damage *= config.base_hit_damage_mult + config.low_hit_damage_mult//Lower chance of a damage buff.
 				if(i == 1) to_chat(user, "<span class='notice'>Your bipod keeps [src] steady!</span>")
```
ninjanomnom: fix this while you're here as well

code/modules/projectiles/updated_projectiles/gun_system.dm
```diff
@@ -692,44 +716,57 @@ and you're good to go.
...
+						apply_bullet_effects(projectile_to_fire, user) //We add any damage effects that we need.
+						projectile_to_fire.dir = get_dir(user, M)
+						projectile_to_fire.distance_travelled = get_dist(user, M)
+						simulate_recoil(1, user)
+
+						if(projectile_to_fire.ammo.bonus_projectiles_amount)
+							var/obj/item/projectile/BP
+							for(var/i = 1 to projectile_to_fire.ammo.bonus_projectiles_amount)
+								BP = rnew(/obj/item/projectile, M.loc)
+								BP.generate_bullet(ammo_list	[projectile_to_fire.ammo.bonus_projectiles_type])
```
ninjanomnom: Mistaken space?

code/modules/projectiles/updated_projectiles/gun_attachables.dm
```diff
@@ -267,27 +270,27 @@ obj/item/attachable/attack_hand(var/mob/user as mob)
...
+			S.remove_from_storage(src)
+		if(loc == user)
+			user.drop_inv_item_on_ground(src)
+		var/obj/item/weapon/combat_knife/F = new(src.loc)
+		user.put_in_hands(F) //This proc tries right, left, then drops it all-in-one.
+		if(F.loc != user) //It ended up on the floor, put it whereever the old flashlight is.
+			F.loc = src.loc
+		cdel(src) //Delete da old bayonet
+	else
+		. = ..()
```
ninjanomnom: If it's the last line `return ..()`

code/modules/projectiles/updated_projectiles/ammo_datums.dm
```diff
@@ -712,159 +817,158 @@
...
-				if(M:fire_immune) continue
+/datum/ammo/rocket/wp/drop_flame(turf/T)
+	if(!istype(T)) return
+	smoke.set_up(1, T)
+	smoke.start()
+	if(locate(/obj/flamer_fire) in T) return
+	new /obj/flamer_fire(T, pick(15, 20, 25, 30))
+	for(var/mob/living/carbon/M in range(3, T))
+		if(istype(M,/mob/living/carbon/Xenomorph))
+			if(M:fire_immune) continue
```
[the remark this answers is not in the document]
ninjanomnom: Also single line control statement

PR #150, opened 2018-08-19
Carrier Numeracy Fix

Incomptinence:
Makes carrier a tier 2 instead of a secret tier 3.

Stops carrier blocking t3 slots despite not checking t3 numbers when evolving itself and being generally out of line with the hivelord and what tier hive status tells you the carrier is in game.

one

TWO

three


Iatots:
I'd just like to say that every old/decent CM player knows this, it's to balance hugger combat. After a whole lot of updates hugger combat is manageable for marines now, and carrier isn't such a strong pick anymore. You used to see about 3 carriers per round at one point last year, now it's usually just one outside peak hours.

Incomptinence:
Even if that is the intent fairly sure carriers can still evolve if t3 is full but t2 is open making it just a tacky way to handle it.

PR #158, opened 2018-08-20
A few minor changes to the Almayer, for decoration and functionality

francinum:
-Ports the Tomb of the Unknown Employee

-Gives the kitchen actual flour
-They also get the randomized barsign


francinum:
The name of arc is up in the air.

Rohesie:
I see no issues here and will merge it in 24 hours unless there's serious opposition.

Rohesie:
The PR is stale and the tomb has been added by another.

PR #159, opened 2018-08-20
Light me up inside (Adds CAS lighting flares)

F-OS:
This PR adds lighting flares for CAS runs. These flares are launched from the minirocket pod and create a pitiful explosion, but create a ton of light. One 6 pack of them costs 250 points.


F-OS:
reeeeeeeeeeeeeeee

code/modules/cm_marines/dropship_ammo.dm
```diff
@@ -348,4 +348,43 @@
...
\ No newline at end of file
+					new/obj/flamer_fire(T)
+
+/obj/structure/ship_ammo/minirocket/illumination
+	name = "illumination rocket-launched flare stack"
+	desc = "A pack of laser guided mini rockets, each loaded with a payload of white-star illuminant and a parachute, while extremely ineffective at damaging the enemy, it is very effective at lighting the battlefield so marines can damage the enemy."
+	icon_state = "minirocket_ilm"
+	point_cost = 250
+
+	detonate_on(turf/impact)
```
ninjanomnom: No relative pathing

code/modules/cm_marines/dropship_ammo.dm
```diff
@@ -348,4 +348,43 @@
...
+
+/obj/item/device/flashlight/flare/on/cas
+	name = "illumination flare"
+	desc = "Report this if you actually see this FUCK"
+	icon_state = "" //No sprite
+	invisibility = 101
+	mouse_opacity = 0
+	brightness_on = 8 //Magnesium/sodium fires (White star) really are bright
+
+	New()
```
ninjanomnom: Same as above

code/modules/cm_marines/dropship_ammo.dm
```diff
@@ -348,4 +348,43 @@
...
+		..()
+		var/turf/T = get_turf(src)
+		fuel = rand(1000, 1100) // About the same burn time as a flare, considering it requires it's own CAS run.
+		T.visible_message("<span class='warning'>You see a flare go up in the sky!</span>")
+		playsound(T, 'sound/weapons/gun_flare.ogg', 50, 1, 4) // stolen from the mortar i'm not even sorry
+		spawn(10) // Sloppy i know but the only way to fix this.
+			update_brightness()
```
ninjanomnom: I remember you saying earlier you found out the smoke is what was causing the lighting issues. Why are you then having the lighting update happen outside the thing causing the need for an update?

F-OS: I found out that having the lighting update again afterwards is a good way to fix many bugs with it. Not just the smoke bug.

F-OS: Outside the scope of this PR, the issues are with CM's particle system and how it treats turfs when it comes to explosions.

ninjanomnom: It's not outside the scope to fix issues that are causing your feature to not work properly

F-OS: But my feature has been tested to work properly when using the delays.

ninjanomnom: Doesn't matter, working code != good code

PR #162, opened 2018-08-20
Makes revolvers fit in sidearm pouches

### LINKED ISSUE #157: Sidearm pouch doesnt carry all sidearms
I noticed this when my sidearm pouch couldn't carry the m44. This is because the sidearm pouch has the line: can_hold = list("/obj/item/weapon/gun/pistol"). As a result items that make sense as sidearms will not fit if they have a different category (such as /obj/item/weapon/gun/revolver). 

There is the M276 belt designed to hold the m44 (under /obj/item/storage/belt/gun/m44) but nobodys going to use a belt slot for a fucking sidearm, nor is it good practice to exclude obvious sidearm weapons from the sidearm pouch. I suggest either A: The sidearm pouch is allowed to carry all items that we think of as sidearms or B: The sidearm pouch is renamed to the pistol pouch. The second option is not really ideal because the code makes few distinctions between revolvers and pistols, often referring to both interchangebly.

JamesDenholm:
Fixes #157 

Really, this is only a temporary fix until we get proper [revolver across-the-back slings](https://www.youtube.com/watch?v=H6cDTVKHsek), of course.

### DIFF
--- a/code/game/objects/items/storage/pouch.dm
+++ b/code/game/objects/items/storage/pouch.dm
@@ -123,7 +123,8 @@
 	desc = "It can contain a pistol. Useful for emergencies."
 	icon_state = "pistol"
 	max_w_class = 3
-	can_hold = list("/obj/item/weapon/gun/pistol")
+	can_hold = list("/obj/item/weapon/gun/pistol",
+					"/obj/item/weapon/gun/revolver")
 	draw_mode = 1
 
 


Wilchenx:
there is a dedicated belt holster for the revolver

JamesDenholm:
@Wilchenx : Refer discussion in #157

code/game/objects/items/storage/pouch.dm
```diff
@@ -123,7 +123,8 @@
 	desc = "It can contain a pistol. Useful for emergencies."
 	icon_state = "pistol"
 	max_w_class = 3
-	can_hold = list("/obj/item/weapon/gun/pistol")
+	can_hold = list("/obj/item/weapon/gun/pistol",
+					"/obj/item/weapon/gun/revolver")
```
Wilchenx: why does this need to be a another line

JamesDenholm: Because basic readability? It's not my fault github has tabs = 8 spaces

ninjanomnom: Multiline lists are fine

PR #164, opened 2018-08-20
changes ooc hotkey from F2 to O

factoryman942:


factoryman942:
wew that's more lines fucked with than should be and I've probably done something horribly wrong haven't I

LetterN:
already did this with #105

LetterN:
also for the amount of lines, it seems like byond updated and made it so that you don't use CAPITAL letters for macro stuff, it's normal

PR #166, opened 2018-08-20
Makes marine weapons spawn with a flashlight

Alterist:
Marine weapons (M4A3, M44, M39, M37A2, M41A Mk2) now spawn with flashlights attached. This is done with the aim of reducing marine prep time, because there's no reason why you wouldn't want a flashlight on your gun if you can't get anything else


subject217:
Just going to lead to a billion weapon lights lying around everywhere. They're already readily available if you want one, and it *really* does not take much time to put them on.

ninjanomnom:
People can put flashlights on if they want, not everything has to be streamlined like this.

PR #168, opened 2018-08-20
Hypervene Recipe - Part of Big Chem Update

Surrealaser:
Added recipe for Hypervene, a purge chem I've submitted a pull request for.

PR #169, opened 2018-08-20
Big Chem Update

Surrealaser:
Adds Hypervene due to a lack of a devoted purge chem so Medics/Doctors don't have to do silly shit like combining smoke/potassium and water in their patient.

Unnerfed Hyperzine; it's not as effective as Ultrazine, metabolizes much faster and has a chance of incurring heart damage due to its relative commonality.

Allows Quickclot and Peridaxon to gradually heal IB and Organ damage respectively in cryotubes (assuming they work).

Makes more meds scannable that should be.

### DIFF
--- a/code/modules/reagents/chemistry_reagents/medical.dm
+++ b/code/modules/reagents/chemistry_reagents/medical.dm
@@ -57,10 +57,10 @@
 		. = ..()
 		for(var/datum/reagent/R in M.reagents.reagent_list)
 			if(R != src)
-				M.reagents.remove_reagent(R.id,5 * REM)
+				M.reagents.remove_reagent(R.id,4 * REM)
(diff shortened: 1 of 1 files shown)


code/modules/reagents/chemistry_reagents/medical.dm
```diff
@@ -128,9 +161,11 @@
 	description = "An effective and very addictive painkiller."
 	reagent_state = LIQUID
 	color = "#C805DC"
-	custom_metabolism = 0.25 // Lasts 10 minutes for 15 units
+	scannable = 1
```
JamesDenholm: No need to define scannable twice

Surrealaser:
Related: https://github.com/ColonialMarines-Mirror/ColonialMarines-2018/pull/168

PR #171, opened 2018-08-20
Make autoinjectors viable again

JamesDenholm:
Surrealistik was complaining on discord so I did a thing

Autoinjectors now take up half as much width (2.0 -> 1.0. The coloured bands are still visible... barely) and the medic's autoinjector pouch now takes 8 (EIGHT) autoinjectors, up from 4 (four...).

You might say "hey, eight is a lot". But that's double four, I didn't invent how math works.

I wonder at what point these PRs are gonna be considered spam.


ninjanomnom:
We've already had significant buffs to medical treatment as a result of skills, let that sit a while first before we go off buffing it elsewhere.

PR #177, opened 2018-08-21
Removing a warning for survivors to not linger around the LZ

F-OS:
This removes the "Don't linger around the LZ" Warning for survivors.

### DIFF
--- a/code/game/gamemodes/cm_initialize.dm
+++ b/code/game/gamemodes/cm_initialize.dm
@@ -649,7 +649,6 @@ datum/game_mode/proc/initialize_special_clamps()
 			else
 				to_chat(H, "\blue You are a survivor of the attack on the colony. You worked or lived in the archaeology colony, and managed to avoid the alien attacks...until now.")
 		to_chat(H, "\blue You are fully aware of the xenomorph threat and are able to use this knowledge as you see fit.")
-		to_chat(H, "\blue You are NOT aware of the marines or their intentions, and lingering around arrival zones will get you survivor-banned.")
 	return 1
 
 /datum/game_mode/proc/tell_survivor_story()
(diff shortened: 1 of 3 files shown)


JamesDenholm:
I'm pretty sure this is still somewhat policy?

JJRcop:
On the /tg/station server, we don't have a policy for golems to follow their creators, that rule is given to them in text similar to this, so this is equivalent. If you want to remove this, I don't think claiming it isn't a part of policy is a valid reason, you should remove it because you want it removed.

Kingtrin:
I don't see the similarity to golems at all. We allow marines to know fully about the xenos before landing, so it would stand to reason that survivors should have knowledge of the marines as well. Furthermore, its idiotic logic to send out a distress signal and not expect any sort of response from the USMC. Even in IC a survivors only real hope would be help arriving at some point. This policy makes zero sense even on CM.

JJRcop:
@Kingtrin
I am not trying to make comparisons between survivors and golems in any meaningful way, except that they are given rules to follow by *the game* that are not explicitly a part of some policy made externally by the administration and written down in the Rules page, just like survivors.

My argument is strictly that removing this based off of "policy" is not a valid reason, because policy doesn't apply here. Removing it because it doesn't make sense to you, you don't like it, or you don't feel that it fits with the aim of DMCA, are valid reasons. We do have Silicon policy, and Security policy, written down in the Rules page on the wiki. These are rules not given to players by *the game*, but are given to them by the administration. There is no Survivor policy written in the DMCA Rules page on the wiki, thus, not fitting with policy is not a valid reason.

I hope this clarifies my argument.

F-OS:
@JJRcop To clarify, i have changed the reason for this PR to i don't feel it fits with the roleplay level here.

PR #180, opened 2018-08-21
Removes the dumb restrictions on first aid kits being used to carry surgical tools.

F-OS:
First aid kits may now be used to carry field surgery supplies, that restriction was dumb and i feel this allows for more interesting forward doctor play.

### DIFF
--- a/code/game/objects/items/storage/firstaid.dm
+++ b/code/game/objects/items/storage/firstaid.dm
@@ -14,7 +14,7 @@
 	throw_speed = 2
 	throw_range = 8
 	cant_hold = list("/obj/item/ammo_magazine",
-							"/obj/item/explosive/grenade") //to prevent powergaming.
+							"/obj/item/explosive/grenade") 
 	var/empty = 0 //whether the kit starts empty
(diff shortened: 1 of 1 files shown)


Kingtrin:
As a field doctor I typically would drag the entire tray in a barrier crate box and leave it on the ground next to my roller bed and anesthetic. The first aid kits I think may be too small to carry all the relevant surgical tools at once but this was a dumb restriction in the first place.

code/game/objects/items/storage/firstaid.dm
```diff
@@ -14,8 +14,7 @@
 	throw_speed = 2
 	throw_range = 8
 	cant_hold = list("/obj/item/ammo_magazine",
-							"/obj/item/explosive/grenade",
-							"/obj/item/tool/surgery") //to prevent powergaming.
+							"/obj/item/explosive/grenade") //to prevent powergaming.
```
ninjanomnom: remove the comment

Iatots:
FWIW I like the way it is now, and trays can't disappear in a dead doctor's pocket.

PR #188, opened 2018-08-22
Destructible/Dismantle-able Window Frames

Surrealaser:
Per title.

Window frames can be damaged, including by heat, projectiles and explosions, and can be repaired with a welder.

Window frames can be dismantled with screwdriver > crowbar > wrench.

### DIFF
--- a/code/game/objects/structures/window_frame.dm
+++ b/code/game/objects/structures/window_frame.dm
@@ -13,7 +13,12 @@
 	var/basestate = "window"
 	var/junction = 0
 	var/reinforced = FALSE
-
+	var/damage = 0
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/window_frame.dm
```diff
@@ -13,7 +13,12 @@
 	var/basestate = "window"
 	var/junction = 0
 	var/reinforced = FALSE
-
+	var/damage = 0
```
ninjanomnom: Same as #189 this should be on /structure. Feel free to combine these 2 prs if you fix it up so structures handle it properly.

PR #189, opened 2018-08-22
Destructible/Dismantle-able Cargo Containers

Surrealaser:
Per title.

Cargo containers can be damaged, including by heat, projectiles and explosions, leaving behind wreckage of 1 metal.

Cargo containers can be dismantled for 2 metal with a welder.


code/game/objects/structures/cargo_container.dm
```diff
@@ -6,10 +6,51 @@
 	bound_width = 32
 	bound_height = 64
 	density = 1
-	var/health = 200
+	var/damage = 0
```
ninjanomnom: This sort of change needs to be done at a lower level like /structure so they can all benefit

PR #191, opened 2018-08-22
highlander

LetterN:
this just makes highlander even more playable


code/__HELPERS/game.dm
```diff
@@ -424,3 +424,7 @@ datum/projectile_data
 /proc/convert_c2k(var/temp)
 	return ((temp + T0C))
 
+/proc/send_to_playing_players(thing) //sends a whatever to all playing players; use instead of to_chat(world, where needed)
```
ninjanomnom: We actually check for `world` in to_chat now instead in tg for the most part

F-OS:
THERE CAN BE ONLY ONE

code/__HELPERS/game.dm
```diff
@@ -424,3 +422,7 @@ datum/projectile_data
...
```
ninjanomnom: Make to_chat support world as a destination instead

code/modules/admin/verbs/onlyone.dm
```diff
@@ -3,33 +3,70 @@
...
-
-		ticker.mode.traitors += H.mind
-		H.mind.special_role = "traitor"
+	switch(alert("Delayed or no?", "PICK ONE!", "NOW!", "Delayed", "Cancel"))
+		if("Cancel")
+			return
+		if("Delayed")
+			for(var/M in player_list)
+				playsound(M, "sound/misc/highlander_delayed.ogg", 40)
+				to_chat(world, "<font size = 4 color=red>Bagpipes begin to blare. You feel Scottish pride coming over you.</span></font>")
```
Rohesie: This is not the only instance that spams one message for every mob hearing it:
https://i.gyazo.com/32e0b63928e00be2d14e795a9fed3118.png
But it's a good example.

Rohesie:
Feel free to ask this to be re-opened when you resume working on it.

PR #193, opened 2018-08-22
Requisition windows are no longer strong enough to sustain a point-blank hit by a Tsar Bomba.

F-OS:
They're still strong however they're no longer unbreakable, i can think of multiple situations where it might be useful to have them breakable.


Rohesie:
This is a good change. Changing the indestructible objects into really tough ones is a philosophy I like and would like to see more, as long as the mapping balance allows.

PR #194, opened 2018-08-22
Fixes jobs list not populating

JamesDenholm:
Unfortunately @LetterN isn't descriptive in their commits so I don't know what the intent of the commit was, BUT it does fix the jobs list not populating

(Yeeeee N I'mma call you out come at me, fight me bro)

But yeah Ninja you're gonna want to test this more because while it works for Extended and Distress Call, idk enough about adminning to be able to work out the full ramifications here

### DIFF
--- a/code/game/jobs/role_authority.dm
+++ b/code/game/jobs/role_authority.dm
@@ -83,7 +83,7 @@ var/global/datum/authority/branch/role/RoleAuthority
 		J = roles_for_mode[i]
 		if(J)
 			L[J.title] = J
-		roles_for_mode = L
+	roles_for_mode = L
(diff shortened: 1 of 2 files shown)


JamesDenholm:
I joke, though, N.

To be clear.

ninjanomnom:
Nah I don't need to test this sort of thing, the structure shouldn't have fundamentally changed from the first commit other than the change to subtypesof and this changes the code back to how it was just with the relative pathing fixed.

JamesDenholm:
gosh DANGIT ninja

LetterN:
i hate the tab key

PR #195, opened 2018-08-22
Replaces CM anti-powergame toxin chems with weaker versions of their reactants.

F-OS:
Right, so CM added some chems that are intended to prevent people from powergaming, for instance, blending hyperzine and peri creates toxin, same with paracetamol and tramadol, same with synaptizine and antitoxin, this PR removes those effects and makes the reactions simply make a weaker version of the chem, for instance, hyper+peri=hypozine, this gives you a very minor speedboost but without organ damage, para+tram=percocet, this reduces pain a bit more then para but way less then tramadol, and 
synap+antitox=Levoamphetamine, this gives minor stun reduction and drowsiness reduction but without the benefits of removing stuff like LSD.

### DIFF
--- a/code/modules/reagents/chemistry_reactions/other.dm
+++ b/code/modules/reagents/chemistry_reactions/other.dm
@@ -49,14 +49,14 @@
 
 
 /datum/chemical_reaction/hypozine
-	name = "hypozine"
+	name = "Hypozine"
 	id = "hypozine"
 	result = "hypozine"
 	required_reagents = list("hyperzine" = 1, "peridaxon" = 1)
 	result_amount = 2
 
 /datum/chemical_reaction/percocet
-	name = "percocet"
+	name = "Percocet"
 	id = "percocet"
 	result = "percocet"
 	required_reagents = list("paracetamol" = 1, "tramadol" = 1)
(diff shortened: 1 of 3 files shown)


JamesDenholm:
_shrug_

This still feels very unnecessary to me

Surrealaser:
As I said in Discord, I don't think this is necessary; the existing system works pretty well, so I'm not in support, but I'm also willing to go along with the majority on this.

Iatots:
So in essence you introduce minor boosts free of downsides that can be stacked with the good ones?

F-OS:
They can't be stacked.

ninjanomnom:
I don't see a particularly good reason for this but I'm not too attached to the old system. I'll close this in 24h or when discussion about this dies down here, whichever comes later.

Sifaus:
I like this one, I think in general i'd really like if the chem/research/etc in DMCA is consistent with the TG but i have no clue how you'd make it balanced.

Rohesie:
The current medical and chem system requires skill. Powergaming is viable, but there are risks. This basically eliminates the risks and makes it less challenging.

I'm very against dumbing down the chem system, which is one of the best things CM has.

JamesDenholm:
I just want to re-iterate my earlier comment, that this isn't necessary -- these aren't chem mixes you see on the field.

Given there's an outstanding chem re-balance, let's see the impact of that, first.

F-OS:
Right, thanks.

PR #197, opened 2018-08-23
Selling xeno corspes feature

GLACoding:
Xenos can be now be sold for requisition points, the number of points depends currently on the caste, although the points for individual types of xenos could be changed.

Powerloaders can also pick up dead xenos on help intent.

This PR aims to give supply personnel a bit more to do in the mid-game, trying to secure dead xenos planet side to sell.
It will also reward the marines for getting kills, and allow marines to play less rush style, as they should keep pace with how strong the aliens get late game, as long as they manage to secure their kills.

Feel free to suggest new numbers for the xeno sell price.

Credits to @hamzah-hayat for helping with the supply shuttle code.


Rohesie:
Good feature, but I'd tweak the values:
1 for larva or removed embryo
5 for T1s
10 for T2s
15 for hivelord/carrier
30 for T3s
60 for queen

Do remember xenos die a lot. Even with the points I'm suggesting it may break balance quite a bit. We'd have to tweak how many points req gets over time and at start to compensate.

GLACoding:
I really don't want to make T1s only 5, they'll only be worth a 1/4 of the cheapest possible crate, and considering the xenos need to be hauled all the way from planet side they should be somewhat worthwhile at least.

Agreed on toning down the numbers in general though.

Rohesie:
Runners are dime a dozen, they die in waves.

FantasticFwoosh:
May i suggest re-utilising the 'fulton' feature from /tg/ to deliver them (and other valuables) back? It seems like a convenience to drag large xenos back manually even with having to bring power loaders & maint techs with nothing better to do down planetside. 

*cough cough, since the lore is becoming homogenized, it'd be a good in-universe explanation behind the history of the miner tech without using bluespace.*

factoryman942:
Have beno corpses be able to be brought back via medevac stretcher, maybe? Then at some point in the far future, if research is to be made, fultons can be an upgrade to the stretcher or something

Rohesie:
Let's keep this simple for now, see if doesn't impact the balance too much, and check how much it's actually used. Then we can figure forms of easier corpse transport, if needed.

iskyp:
\>lone infiltrator in xeno hive
\>manage to use your skill in close quarters combat to quickly take down a runner
\>extract him
>[Supply] RO Kazuhira Miller says "He's coming too? Roger that."

PR #198, opened 2018-08-23
Dead Xenos can be buckled to roller beds

GLACoding:
Added for #197 as a way to reduce drag delay. Mostly to make dragging dead xenos from planet side less tedious.

### DIFF
--- a/code/game/objects/objs.dm
+++ b/code/game/objects/objs.dm
@@ -13,7 +13,7 @@
 	var/mob/living/buckled_mob
 	var/buckle_lying = FALSE //Is the mob buckled in a lying position
 	var/can_buckle = FALSE
-	
+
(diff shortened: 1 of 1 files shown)


Sifaus:
Wouldn't it be better (thematic?) to make it so that Xenos can be put inside big boxes/crates? Like, carrying a huge ravager on a human-size roller bed would look kinda funny. Although, i think in general, finding a way to make xeno dragging less tedious would be fine, if #197 is merged.

Rohesie:
Yes, dragging a dead queen on a roller bed would be pretty odd visually. Also if the corpses start having value the cost of dragging them is justified, no need to make it super-easy to carry 20 T3s into req.

A dead queen is worth an HPR after all.

PR #200, opened 2018-08-23
Synths dewhitelist

Rohesie:
* Synths are no longer whitelisted as species or job.
* Stun reduction removed.
* Melee base damage equal to that of a human.
* Leadership and CQC skills removed.
* Melee skill kept for old model synths (they are slower but sturdier and hit a bit harder, not half as strong as before), but not for the faster and newer one.

### DIFF
--- a/code/datums/skills.dm
+++ b/code/datums/skills.dm
@@ -129,16 +129,13 @@
 
 /datum/skills/synthetic
 	name = "Synthetic"
-	cqc = SKILL_CQC_MASTER
 	engineer = SKILL_ENGINEER_MT
 	construction = SKILL_CONSTRUCTION_MASTER
 	firearms = SKILL_FIREARMS_TRAINED
 	smartgun = SKILL_SMART_TRAINED
 	spec_weapons = SKILL_SPEC_TRAINED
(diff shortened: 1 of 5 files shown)


Kingtrin:
Thanks for looking into this and making changes. We had a good discussion on the forums and discord. I agree with these for the most part but I think that having metal hands should make you punch at least slightly harder. The melee skill is done in increments of 30% per level (what the fuck) so its really not fair to use this. CQC offers some undescribed amount of unarmed damage increase but I assume you removed it for a reason. I'm not sure what the best answer to this is but having base stats means its easier to balance the game around them.

Due to the ongoing stun rework its safer to remove stun resist for now so I also think this is a good change. Above all, I happiest about our new admin: Rohesie! congratulations on editing yourself into a define file!

I hope synths get some more love in the future :)

Kingtrin:
Addenum - There is still a whitelist flag in code/game/jobs/job/civilians.dm for the synths

flags_whitelist = WHITELIST_SYNTHETIC 

Will this cause issues?

Rohesie:
It won't cause issues, no.

PR #204, opened 2018-08-24
Gas Mask Now Fits In Suit Slot

Surrealaser:
Feel this is a balanced alternative to the current state of the gas mask: allows PFCs and others that tend not to carry packs to carry a gas mask, but the delay of taking it out of your suit slot and putting it on should be significant enough so as to not be overtuned, especially in tandem with the vision impairment.


FantasticFwoosh:
Finally a solution to my frustrated suit slot clicking, the thing's take up _SOOO_ much space in my inventory which is really not what i need as a engineer, nor the lack of vision from storing it on my face.

Issue #24, opened 2018-07-31
Fix admin logs so they dont show every bullet being shot


DoctorPork:
It's a cascading nightmare.

CosmicScientist:
Dual wield SMGs you say?

JJRcop:
https://github.com/ColonialMarines-Mirror/ColonialMarines-2018/blob/fe26b64bc44add91c2a6f31422c94832148bd025/code/modules/projectiles/updated_projectiles/projectile.dm#L761-L773

remove the msg_admin_attack messages and msg_admin_ff

iskyp:
[literally this](https://www.youtube.com/watch?v=nYibgPhqQhY)

Rohesie:
There is a verb to toggle attacks and one to toggle FF.

PR #209, opened 2018-08-24
Logging port

### LINKED ISSUE #155: Logging is terrible!
You access one single log for all logging types, you can't access logs for individual players, and the logging is grouped per-day rather than per-round. On top of that, in-game, you cannot access logs past the current day.

Our logging tools should be ported.
@rustledjimm 
@francinum

JJRcop:
This ports the /tg/station logging system, including the new refactor. Each player has individual logs, logs are separated per-round rather than per-day.

This should look familiar

Closes #155


Iatots:
Does it fix #24 and #155 ?

JJRcop:
It doesn't fix #24 but it does fix #155. #24 should be simple to fix

francinum:
#24's core issue is attack logs being turned on for admins by default.

JJRcop:
By the way, I didn't port everything, like for example tcomms logs, just most things

Rohesie:
This has been testmerged and worked fine.
I haven't seen any glaring issues with the code, so after @ninjanomnom gives the green light this will be merged.

JJRcop:
After a few days i'm going to do another ```world.log <<``` pass to catch any new changes.

PR #210, opened 2018-08-25
Follow humans now only shows living, non-SSD humans.

F-OS:


F-OS:
👏 I 👏 AM 👏 SHIT 👏 AT 👏 GIT

Rohesie:
This is tested and works as intended. I see no issues with it.

PR #219, opened 2018-08-26
Tanks can now break windows and window frames.

F-OS:
This allows better combat progression and reduces slog.


Rohesie:
This was on my TO DO list, along with other tank fixes. I'll leave it open for 24 hours and merge it if it has no serious opposition. Already implemented in CM, even.

ninjanomnom:
Just merge it when you want to merge it @Rohesie

PR #221, opened 2018-08-26
Adds Polaris' towels... with a twist!  ALSO belts can now be worn without jumpsuits/uniforms ALSO no more starting backpack/satchel for people that wake from cryosleep.

Iatots:
Original PR from Yoshax with sprites from Pawn, this PR only uses the sprites with a 2 sprite addition.

Marines that spawn out of cryo will do so wearing a towel now.


Rohesie:
No WEAR_JACKET towels for females?

code/game/objects/items.dm
```diff
@@ -371,7 +371,7 @@ cases. Override_icon_state should be a list.*/
 						to_chat(H, "<span class='warning'>You need a jumpsuit before you can attach this [name].</span>")
 					return 0
 				if(!(flags_equip_slot & SLOT_WAIST))
-					return
+					return 0
```
Rohesie: Do a `return FALSE` instead.

Iatots: The whole file needs it. Wouldn't want to mess with a later regex scrub.

Iatots:
For female discrimination, should I use an if check inside the generate_wearable proc?

code/game/objects/items.dm
Rohesie: Feel free to do it to the whole file.

Rohesie:
There's also the `Expr ? TrueExpr : FalseExpr` syntax. Works like an `if (Expr) then TrueExpr else FalseExpr`
Check the "? operator" in the BYOND documentation for more info. You can press F1 in Dream Maker for quick access to it.

code/game/jobs/job/marines.dm
```diff
@@ -21,6 +21,17 @@
 		. = ..()
 		if(flags_startup_parameters & ROLE_ADD_TO_MODE) H.nutrition = rand(60,250) //Start hungry for the default marine.
 
+	generate_wearable_equipment()
```
Rohesie: Check the "All BYOND paths must contain the full path" in the guide to contributing: https://github.com/ColonialMarines-Mirror/ColonialMarines-2018/blob/master/CONTRIBUTING.md

code/game/jobs/job/marines.dm
```diff
@@ -11,16 +11,25 @@
...
+		. = "You have been assigned to: <b><font size=3 color=[squad_colors[H.assigned_squad.color]]>[lowertext(H.assigned_squad.name)] squad</font></b>.[flags_startup_parameters & ROLE_ADD_TO_MODE ? " Make your way to the cafeteria for some post-cryosleep chow, and then get equipped in your squad's prep room." : ""]"
 
-	generate_entry_conditions(mob/living/carbon/human/H)
-		. = ..()
-		if(flags_startup_parameters & ROLE_ADD_TO_MODE) H.nutrition = rand(60,250) //Start hungry for the default marine.
+/datum/job/marine/generate_entry_conditions(mob/living/carbon/human/H)
+	. = ..()
+	if(flags_startup_parameters & ROLE_ADD_TO_MODE) H.nutrition = rand(60,250) //Start hungry for the default marine.
 
+/datum/job/marine/generate_wearable_equipment()
```
Rohesie: You'll might want to add it to a list `+=` instead of overriding `=`.

code/game/jobs/job/marines.dm
```diff
@@ -11,16 +11,25 @@
...
+	if(MALE)
+		. = list(
```
Rohesie: You'll might want to add it to a list `+=` instead of overriding `=`.

Iatots: I think i need to put += on every other instance actaully, since this is the parent.

code/game/jobs/job/marines.dm
```diff
@@ -11,16 +11,25 @@
...
+/datum/job/marine/generate_entry_message(mob/living/carbon/human/H)
+	if(H.assigned_squad)
+		. = "You have been assigned to: <b><font size=3 color=[squad_colors[H.assigned_squad.color]]>[lowertext(H.assigned_squad.name)] squad</font></b>.[flags_startup_parameters & ROLE_ADD_TO_MODE ? " Make your way to the cafeteria for some post-cryosleep chow, and then get equipped in your squad's prep room." : ""]"
 
-	generate_entry_conditions(mob/living/carbon/human/H)
-		. = ..()
-		if(flags_startup_parameters & ROLE_ADD_TO_MODE) H.nutrition = rand(60,250) //Start hungry for the default marine.
+/datum/job/marine/generate_entry_conditions(mob/living/carbon/human/H)
+	. = ..()
+	if(flags_startup_parameters & ROLE_ADD_TO_MODE) H.nutrition = rand(60,250) //Start hungry for the default marine.
```
F-OS: This is pet peeve #1 for me.

code/game/jobs/job/marines.dm
Rohesie: Procs get overridden when redefined for more specific objects.
`/proc/do()` won't be called if you have `/proc/mob/do()`, if `do()` is being called by the mob object, and it contains no `..()`

There's several ways you could do this.

Rohesie: `/datum/job/marine/leader/generate_wearable_equipment()`, for example, could use a `. = ..()` at the beginning to add the towels to the list it will output.
Does the logic of this make sense to you?
You have two lists that will deal with each job, one for everyone and one for each specific case. Only one of them will be called, but that one can call the other. You want both lists to be output, so the one which will be called (the job-specific one) has to call the general one, get that list, and then add the specific contents to it before outputting them.

code/game/jobs/job/marines.dm
```diff
@@ -11,16 +11,26 @@
...
+	. = ..()
+	if(flags_startup_parameters & ROLE_ADD_TO_MODE) 
+		H.nutrition = rand(60,250) //Start hungry for the default marine.
 
+/datum/job/marine/generate_wearable_equipment(mob/living/carbon/human/H)
+	if (H.gender == MALE)
+		. = list(
+			WEAR_WAIST = /obj/item/weapon/twohanded/towel
+			)
+	if (H.gender == FEMALE)
```
Rohesie: `else if` here?

code/game/jobs/job/marines.dm
```diff
@@ -35,13 +45,9 @@
 	flags_startup_parameters = ROLE_ADD_TO_DEFAULT|ROLE_ADD_TO_MODE|ROLE_ADD_TO_SQUAD
 	skills_type = /datum/skills/SL
 
-	generate_wearable_equipment()
```
Rohesie: Don't delete the specific equipments. Just add a `..()` at the start of it and then add to the list.

Iatots: I did that for the specialist to spawn with his specrag, but I chose to remove the backpack/satchel because marines already get one in their vendor, and sleeping with one is kinda dumb. Both on CM and DMCA people are ok to do away with them.

PR #223, opened 2018-08-27
Slight rebalance to make field surgery more viable

F-OS:
🆑 
NT no longer sprays their roller beds with Staph.
Every bottle of oxycontin you buy now has a free dose of muscle relaxants added in FREE!
Roller beds now have tighter straps.
🆑

### DIFF
--- a/code/modules/surgery/surgery.dm
+++ b/code/modules/surgery/surgery.dm
@@ -89,7 +89,7 @@ proc/spread_germs_to_organ(datum/limb/E, mob/living/carbon/human/user)
 		E.germ_level += user.germ_level / 2
 
 	if(locate(/obj/structure/bed/roller, E.owner.loc))
-		E.germ_level += 100
+		E.germ_level += 50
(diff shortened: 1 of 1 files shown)


Rohesie:
This is old and should have been addressed before. I've thought about it, though. In short the problem is that it makes field surgery easier without really providing any trade-off. You won't have many germ problems already if you carry latex gloves, a medical mask, a roller bed and a spray bottle.

I was thinking about making a surgical apron (and you could add other aprons with smaller effects, same with other gloves and masks), so you could be a little more sterile by sacrificing wearing an armor, for example.

No need to implement that, but it's the line we want to follow. Trade-offs.

Rohesie:
Surgeries without proper equipment are supposed to be harder. They are still very viable if you have the tools, however.
A straight buff is not the way. Making a way to build surgery tables or otherwise create difficult and static ideal surgery conditions is more interesting instead. We don't want to make non-field-surgery obsolete.

PR #225, opened 2018-08-27
Updated Med HUD So It Ignores Ghostless Bodies

Surrealaser:
Per title.


Iatots:
This is what DNR is correct?

Surrealaser:
@Iatots Exactly

Rohesie:
QoL update with no drawbacks.
I'll merge in 24 hours unless I see any objections.

PR #226, opened 2018-08-27
Move Orders Improve Human Melee Damage

Surrealaser:
Per title.

Each level of Leadership beyond the first has Move Orders increase melee damage by 33% while it's in effect (max of +99% for the CO).

Bayonet charges now actually worth a shit?

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -148,7 +148,9 @@ Contains most of the procs that are called when a mob is attacked by something
 		visible_message("\red <B>[src] has been [pick(I.attack_verb)] in the [hit_area] with [I.name] by [user]!</B>", null, null, 5)
 	else
 		visible_message("\red <B>[src] has been attacked in the [hit_area] with [I.name] by [user]!</B>", null, null, 5)
-
+	
+	if(command_aura == "move") //Check for move aura; improves melee damage significantly.
+		I.force *= (1 + (command_aura_strength * 0.33)) // CO move orders temporarily double melee damage (guns are probably still better, but improves those PBs).
 	var/armor = run_armor_check(affecting, "melee", "Your armor has protected your [hit_area].", "Your armor has softened hit to your [hit_area].")
 	var/weapon_sharp = is_sharp(I)
 	var/weapon_edge = has_edge(I)


Rohesie:
Double damage may be quite strong in the right conditions, specially when the marines are faster, but there are only 3 people who can give CO-level orders.

Perhaps tie this to an exception so that whoever gives the order gets no benefit from it?

Let's hear comments about this.

Surrealaser:
@Rohesie As I said before, I'm considering reducing it to +5 damage and +20% of base, so CO level orders are more reasonable, without SL orders being too crap.

Rohesie:
My issue is for this to become a support tool, not a rambo one. An exception to the emiter for the leadership bonus would still be desirable.

Surrealaser:
@Rohesie That sounds like you have a problem with how Orders work in general.

Rohesie:
Their current effects are low enough for that not to be a concern. If we start buffing them, added to the fact that they'll be more common due to the orders icon, then that changes.

FantasticFwoosh:
``I *scream for ice cream``

Xenomorphs have all of the pheromones for buffs, I don't see why *scream(ing) before a charge on a cooldown couldn't give a temporary amp up to melee power to make individual charges a bit more powerful unless the xeno evades you with agility.

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -148,7 +148,9 @@ Contains most of the procs that are called when a mob is attacked by something
 		visible_message("\red <B>[src] has been [pick(I.attack_verb)] in the [hit_area] with [I.name] by [user]!</B>", null, null, 5)
 	else
 		visible_message("\red <B>[src] has been attacked in the [hit_area] with [I.name] by [user]!</B>", null, null, 5)
-
+	
+	if(command_aura == "move") //Check for move aura; improves melee damage significantly.
+		I.force *= (1 + (command_aura_strength * 0.33)) // CO move orders temporarily double melee damage (guns are probably still better, but improves those PBs).
```
octareenroon91: `/mob/living/carbon/human/proc/attacked_by(var/obj/item/I, var/mob/living/user, var/def_zone)`
Are you sure you're on the right proc?

Do weapons forces routinely get recalculated? If not, then weapons used against humans will grow exponentially in force.

PR #227, opened 2018-08-27
More Scannable Chems + Hypervene

Surrealaser:
Makes more chems scannable.

Adds Hypervene purge chem; peripheral elements relating to the chem will be found in another PR

### DIFF
--- a/code/modules/reagents/chemistry_reagents/medical.dm
+++ b/code/modules/reagents/chemistry_reagents/medical.dm
@@ -11,7 +11,6 @@
 	color = "#C8A5DC" // rgb: 200, 165, 220
 	overdose = REAGENTS_OVERDOSE*2
 	overdose_critical = REAGENTS_OVERDOSE_CRITICAL*2
-
 	scannable = 1
 
 	on_mob_life(mob/living/M, alien)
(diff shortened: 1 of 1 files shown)


code/modules/reagents/chemistry_reagents/medical.dm
```diff
@@ -43,6 +42,52 @@
...
+		M.reagent_shock_modifier += 50 //Significant pain while metabolized.
+		if(ishuman(M))
+			var/mob/living/carbon/human/H = M
+			if(prob(5)) //causes vomiting
+				H.vomit()
+		M.adjustToxLoss(-4 * REM)
+		M.radiation = max(M.radiation-8*REM,0)
+
+	on_overdose(mob/living/M, alien)
+		if(alien == IS_YAUTJA) return
```
F-OS: Snowflake check that perpetuates a snowflake species when there is no reason codewise not to have it.

Surrealaser: @F-OS Hey man, I'm just following the current convention. I'd be more than happy to get rid of Predator exceptions.

F-OS: I understand, sorry. I'll actually submit a PR later removing those snowflake exceptions.

code/modules/reagents/chemistry_reagents/medical.dm
```diff
@@ -43,6 +42,52 @@
...
+	overdose_critical = REAGENTS_OVERDOSE_CRITICAL / 2
+	scannable = 1
+
+	on_mob_life(mob/living/M, alien)
+		. = ..()
+		if(!.) return
+		for(var/datum/reagent/R in M.reagents.reagent_list)
+			if(R != src)
+				M.reagents.remove_reagent(R.id,4 * REM)
+		M.reagent_shock_modifier += 50 //Significant pain while metabolized.
```
Rohesie: Create new defines if necessary, and use stuff like PAIN_REDUCTION_LIGHT instead of numbers here. Magical numbers and all.

code/modules/reagents/chemistry_reagents/medical.dm
```diff
@@ -51,6 +96,7 @@
 	color = "#C8A5DC" // rgb: 200, 165, 220
 	overdose = REAGENTS_OVERDOSE
 	overdose_critical = REAGENTS_OVERDOSE_CRITICAL
+	scannable = 1
```
Rohesie: Minor issue, but could you get me a TRUE?

code/modules/reagents/chemistry_reagents/medical.dm
```diff
@@ -623,14 +671,26 @@
...
-	description = "Used to stabilize internal organs while waiting for surgery. Medicate cautiously."
+	description = "Used to stabilize internal organs while waiting for surgery. Medicate cautiously. Repairs organ damage at cryogenic temperatures."
 	reagent_state = LIQUID
 	color = "#C845DC"
 	overdose = REAGENTS_OVERDOSE/2
 	overdose_critical = REAGENTS_OVERDOSE_CRITICAL/2
 	custom_metabolism = 0.05
 	scannable = 1
 
+	on_mob_life(mob/living/M)
```
Rohesie: Use full paths here. All BYOND paths must contain the full path, and all.

code/modules/reagents/chemistry_reagents/medical.dm
```diff
@@ -623,14 +671,26 @@
...
+		. = ..()
+		if(!.) return
```
Rohesie: Get the return into a new line, not in the same as the if() statement.

code/modules/reagents/chemistry_reagents/medical.dm
```diff
@@ -686,12 +758,13 @@
...
 	overdose = REAGENTS_OVERDOSE/5
 	overdose_critical = REAGENTS_OVERDOSE_CRITICAL/5
+	scannable = 1
 
 	on_mob_life(mob/living/M)
 		. = ..()
 		if(!.) return
 
-		M.reagent_move_delay_modifier -= 0.5
+		M.reagent_move_delay_modifier -= 9
```
Rohesie: What's the advantage of ultrazine over hyperzine in this context?

Surrealaser: Ultrazine provides a better speed bonus, lasts far longer, has a much higher OD ceiling, and no heart damage side effects. Hyperzine but better with no side effects except an addiction that is long deferred.

PR #230, opened 2018-08-27
Clean turfs

LetterN:
cleans up things in code/game/turfs/..


code/game/turfs/floor.dm
```diff
@@ -184,89 +174,90 @@ var/list/wood_icons = list("wood", "wood-broken")
...
+		return
+	if(broken)
+		return
 	broken = TRUE
 	if(is_plasteel_floor())
 		icon_state = "damaged[pick(1, 2, 3, 4, 5)]"
-		broken = 1
-	else if(is_light_floor())
+
+	if(is_light_floor())
```
Rohesie: There are no early returns here and several of the options are mutually-exclusive. A `switch` would also work.

code/game/turfs/floors/map_floors.dm
```diff
@@ -0,0 +1,578 @@
...
+
+/turf/open/mars_dirt
+	name = "dirt"
+	icon = 'icons/turf/bigred.dmi'
+	icon_state = "mars_dirt_1"
+
+/turf/open/mars_dirt/New()
+	..()
+	spawn(10)
+		if (rand(0, 32) && icon_state == "mars_dirt_4")
```
Rohesie: This seems a little unintuitive. One chance in 33 not to change the icon to something else? Without an early return so then it checks again.

Why not a `rand` followed by a `switch` or a `prob` followed by a `pick`?

LetterN: probably for chance thing, instead of picking turf

Rohesie:
Feel free to ask this to be re-opened when you resume working on it.

PR #233, opened 2018-08-27
NOPE STILL RETARDED

Alterist:
Ported from the Russian DMCA

    Fixes shooting at darkness with artillery mod.
    Adds a Tank Fabricator(based on the DS Fab) into the code which has limited points that do not generate. This gives the TC a pointbuy system of sorts. (not placed on the map yet)
    Allows people with engineering skill to see a rough status of the tank's hardpoint modules

This isn't a direct port of the code, which includes some re-balances to module costs/hp and the damage value of the LTB and TOW, among others. I decided to leave the values unchanged for now, and leave the re-balancing up for discussion


chesse20:
well lets put this in the game

PR #234, opened 2018-08-27
Tank Update

Alterist:
Ported from the Russian DMCA

Adds a Tank Fabricator(based on the DS Fab) into the code which has limited points that do not generate. This gives the TC a pointbuy system of sorts. (not placed on the map yet)
Allows people with engineering skill to see a rough status of the tank's hardpoint modules
Gives Tank Crew the required engineering skill needed to see the hardpoint hp states, also allows them to use C4

This isn't a direct port of the code, which includes some re-balances to module costs/hp and the damage value of the LTB and TOW, among others. I decided to leave the values unchanged for now, and leave the re-balancing up for discussion


F-OS:
da comrade

FighterX2500:
Wait, artmod isn't functional yet. You might exclude it, idk why, but update_sight seems not working with it (it doesn't have night vision)

Alterist:
I thought the point of the Artmod var was to make it so that you can shoot at something that isn't lit without the shot veering off?

Iatots:
to use the artillery mod: select support as weapon, click a direction to zoom out, switch back to ranged weapon of choice.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1462,6 +1462,10 @@
 			see_in_dark = 8
 			see_invisible = SEE_INVISIBLE_LEVEL_TWO
 
+		if(artmod_use == 1)
+			see_in_dark = 20
+			see_invisible = SEE_INVISIBLE_MINIMUM
```
Rohesie: Mesons would be more balanced than night-vision.

code/modules/vehicles/cm_armored.dm
```diff
@@ -278,7 +278,19 @@ var/list/TANK_HARDPOINT_OFFSETS = list(
 		if(!HP)
 			to_chat(user, "There is nothing installed on the [i] hardpoint slot.")
 		else
-			to_chat(user, "There is a [HP.health <= 0 ? "broken" : "working"] [HP] installed on the [i] hardpoint slot.")
+			if((user.mind && user.mind.cm_skills && user.mind.cm_skills.engineer >= SKILL_ENGINEER_METAL) || isobserver(user))
```
Rohesie: This could better use a `switch` statement instead of checking every condition, given that only one can be true at a time.

code/modules/vehicles/tankvendor.dm
```diff
@@ -0,0 +1,123 @@
...
+
+/obj/machinery/tank_part_fabricator/Topic(href, href_list)
+	if(..())
+		return
+
+	usr.set_interaction(src)
+	add_fingerprint(usr)
+
+	if(busy)
+		usr << "<span class='warning'>The autolathe is busy. Please wait for completion of previous operation.</span>"
```
Rohesie: Use the `to_chat()` standard.

FantasticFwoosh:
Pity on the points, i wanted to flood das aliums with my gommunistic means of production by destroying xeno lines with T-34's en-mass.

For The Motherland

* Is there any chance that some maps for future use and revision could have a battered and broken tank on them so we can spend points at the tank lathe, some engineer elbow power, probably replacement parts at requisition and have two of them on the field as a side objective after getting it running again?

PR #235, opened 2018-08-28
Defines Maths.dm : Bootleg edition

Ghommie:
As the tin says, this PR (A port from /tg/station, original by Ninjanomnom) deletes _HELPERS/maths and replaces it with _DEFINES/maths. I had no errors and warnings when compiling or any runtiming while testing it locally. (albeit there is the slight chance I may have forgot to commit one or two things due to my sleepy state)

I don't even expect this to be merged, but I want to see if travis even compiles and reviews, ahem.


code/game/gamemodes/cm_initialize.dm
```diff
@@ -89,9 +89,9 @@ Additional game mode variables.
 
 datum/game_mode/proc/initialize_special_clamps()
 	var/ready_players = num_players() // Get all players that have "Ready" selected
-	xeno_starting_num = Clamp((ready_players/7), xeno_required_num, INFINITY) //(n, minimum, maximum)
-	surv_starting_num = Clamp((ready_players/25), 0, 8)
-	merc_starting_num = Clamp((ready_players/3), 1, INFINITY)
+	xeno_starting_num = CLAMP((ready_players/7), xeno_required_num, INFINITY) //(n, minimum, maximum)
+	surv_starting_num = CLAMP((ready_players/25), 0, 8)
+	merc_starting_num = CLAMP((ready_players/3), 1, INFINITY)
```
ninjanomnom: clamp with one of the values being infinity is just min() or max()

code/__HELPERS/game.dm
```diff
@@ -223,7 +223,7 @@
 					. |= M		// Since we're already looping through mobs, why bother using |= ? This only slows things down.
 	return .
 
-#define SIGN(X) ((X<0)?-1:1)
+#define SIGN_CM(X) ((X<0)?-1:1)
```
ninjanomnom: The difference I suppose is this one will say SIGN(0) is 1. Don't have both of these but handle that wherever that functionality was depended on.

code/__DEFINES/maths.dm
```diff
@@ -0,0 +1,210 @@
...
+// Secant
+#define SEC(x) (1 / cos(x))
+
+// Cosecant
+#define CSC(x) (1 / sin(x))
+
+#define ATAN2(x, y) ( !(x) && !(y) ? 0 : (y) >= 0 ? arccos((x) / sqrt((x)*(x) + (y)*(y))) : -arccos((x) / sqrt((x)*(x) + (y)*(y))) )
+
+//Finds nearest integer to x, above or below
+/proc/RoundNearest(x)
```
ninjanomnom: ...This is just a normal rounding function

Ghommie: Oh really, wow. They made this complicated walk around to fancy rounding. How could I have not noticed.

code/__HELPERS/game.dm
Ghommie: I removed SIGN_CM and typed its math on the three only istances of game.dm it was called in, for the time being. Also removed a left over #undef SIGN, cause of many undefined procs errors.

code/game/gamemodes/cm_initialize.dm
Ghommie: Done, min().

code/game/gamemodes/cm_initialize.dm
```diff
@@ -89,9 +89,9 @@ Additional game mode variables.
 
 datum/game_mode/proc/initialize_special_clamps()
 	var/ready_players = num_players() // Get all players that have "Ready" selected
-	xeno_starting_num = CLAMP((ready_players/7), xeno_required_num, INFINITY) //(n, minimum, maximum)
-	surv_starting_num = CLAMP((ready_players/25), 0, 8)
-	merc_starting_num = CLAMP((ready_players/3), 1, INFINITY)
+	xeno_starting_num = min((ready_players/7), xeno_required_num, INFINITY) //(n, minimum, maximum)
+	surv_starting_num = min((ready_players/25), 0, 8)
+	merc_starting_num = min((ready_players/3), 1, INFINITY)
```
ninjanomnom: You do need to remove the infinity. Also this should be max
`CLAMP(x, 1, INFINITY)` is equivalent to `max(x, 1)`

code/modules/cm_marines/shuttle_backend.dm
```diff
@@ -407,8 +407,8 @@ cdel(src)
 		y = C.y_pos
 		C1.x_pos = x*cos(deg) + y*sin(deg)
 		C1.y_pos = y*cos(deg) - x*sin(deg)
-		C1.x_pos = RoundNearest(C.x_pos) //Sometimes you get very close to the right number but off by around 1e-15 and I want integers dammit
-		C1.y_pos = RoundNearest(C.y_pos)
+		C1.x_pos = round(C.x_pos) //Sometimes you get very close to the right number but off by around 1e-15 and I want integers dammit
+		C1.y_pos = round(C.y_pos)
```
ninjanomnom: round without a second argument acts like floor which is probably why some mistaken individual made RoundNearest. `round(x, 1)` is the proper replacement

code/game/gamemodes/cm_initialize.dm
```diff
@@ -89,9 +89,9 @@ Additional game mode variables.
 
 datum/game_mode/proc/initialize_special_clamps()
 	var/ready_players = num_players() // Get all players that have "Ready" selected
-	xeno_starting_num = Clamp((ready_players/7), xeno_required_num, INFINITY) //(n, minimum, maximum)
-	surv_starting_num = Clamp((ready_players/25), 0, 8)
-	merc_starting_num = Clamp((ready_players/3), 1, INFINITY)
+	xeno_starting_num = max((ready_players/7), xeno_required_num) //(n, minimum, maximum)
+	surv_starting_num = max((ready_players/25), 0, 8)
```
ninjanomnom: `CLAMP(ready_players/25, 0, 8)` is a valid clamp

Rohesie:
Tested, @ninjanomnom . Server is not on fire, game is unbroken.

ninjanomnom:
Sounds good to me

PR #237, opened 2018-08-28
Ports APC optimization from tg, APCs are now added to a list on creation, get_apc uses that list, this reduces locate calls and speeds get_apc up.

F-OS:
Tested to compile, will do AB test later.

### DIFF
--- a/code/global.dm
+++ b/code/global.dm
@@ -227,8 +227,6 @@ var/list/AAlarmWireColorToFlag = RandomAAlarmWires()
 var/list/AAlarmIndexToFlag
 var/list/AAlarmIndexToWireColor
 var/list/AAlarmWireColorToIndex
-// APC optimization
-var/static/list/apcs_list = list()
 
 #define SPEED_OF_LIGHT 3e8 //not exact but hey!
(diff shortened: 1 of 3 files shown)


code/global.dm
```diff
@@ -227,6 +227,8 @@ var/list/AAlarmWireColorToFlag = RandomAAlarmWires()
 var/list/AAlarmIndexToFlag
 var/list/AAlarmIndexToWireColor
 var/list/AAlarmWireColorToIndex
+// APC optimization
+var/static/list/apcs_list = list()
```
ninjanomnom: It doesn't need to be static

code/modules/power/power.dm
```diff
@@ -279,11 +279,9 @@
 	return null
 
 /area/proc/get_apc()
-	for(var/area/RA in src.related)
-		var/obj/machinery/power/apc/FINDME = locate() in RA
-		if (FINDME)
-			return FINDME
-
+	for(var/obj/machinery/power/apc/APC in apcs_list)
```
ninjanomnom: Don't use full caps names for non defines

PR #238, opened 2018-08-28
Very very slightly optimizes can_see, it no longer runs one more time then needed.

F-OS:
Profile results (total time)
Proc Name                                                                  Self CPU    Total CPU    Real Time        Calls
----------------------------------------------------------------------    ---------    ---------    ---------    
/proc/can_see_b                                                               0.063        0.063        0.073        100001 (Old version)
/proc/can_see_a                                                               0.062        0.062        0.070        100001 (New version)

I realize this is a very tiny difference but i'm a performance whore so idk.

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -752,7 +752,7 @@ proc/anim(turf/location,atom/target,a_icon,a_icon_state as text,flick_anim as te
 	var/turf/target_turf = get_turf(target)
 	var/steps = 1
 	if(current == target_turf)
-  		return TRUE
+		return TRUE
 	if(steps > length)
 		return FALSE
 	current = get_step_towards(current, target_turf)
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -747,21 +747,24 @@ proc/anim(turf/location,atom/target,a_icon,a_icon_state as text,flick_anim as te
 	return toReturn
 
 //Step-towards method of determining whether one atom can see another. Similar to viewers()
-/proc/can_see(var/atom/source, var/atom/target, var/length=5) // I couldnt be arsed to do actual raycasting :I This is horribly inaccurate.
+/proc/can_see(atom/source, atom/target, length=5) // I couldnt be arsed to do actual raycasting :I This is horribly inaccurate.
 	var/turf/current = get_turf(source)
 	var/turf/target_turf = get_turf(target)
-	var/steps = 0
-
-	while(current != target_turf)
-		if(steps > length) return 0
-		if(current.opacity) return 0
-		for(var/atom/A in current)
-			if(A.opacity) return 0
+	var/steps = 1
+	if(current != target_turf)
 		current = get_step_towards(current, target_turf)
-		steps++
-
+		while(current != target_turf)
+			if(steps > length)
+				return 0
+			if(current.opacity)
+				return 0
+			for(var/thing in current)
+				var/atom/A = thing
+				if(A.opacity)
+					return 0
+			current = get_step_towards(current, target_turf)
+			steps++
 	return 1
-
 /proc/is_blocked_turf(var/turf/T)
 	var/cant_pass = 0
 	if(T.density) cant_pass = 1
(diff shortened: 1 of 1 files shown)


code/__HELPERS/unsorted.dm
```diff
@@ -747,21 +747,24 @@ proc/anim(turf/location,atom/target,a_icon,a_icon_state as text,flick_anim as te
...
 	var/turf/target_turf = get_turf(target)
-	var/steps = 0
-
-	while(current != target_turf)
-		if(steps > length) return 0
-		if(current.opacity) return 0
-		for(var/atom/A in current)
-			if(A.opacity) return 0
+	var/steps = 1
+	if(current != target_turf)
```
ninjanomnom: To simplify reading this and reduce indentation just do
```
if(current == target_turf)
  return TRUE
...
```

code/__HELPERS/unsorted.dm
```diff
@@ -747,21 +747,24 @@ proc/anim(turf/location,atom/target,a_icon,a_icon_state as text,flick_anim as te
...
 		current = get_step_towards(current, target_turf)
-		steps++
-
+		while(current != target_turf)
+			if(steps > length)
+				return 0
```
ninjanomnom: FALSE

code/__HELPERS/unsorted.dm
```diff
@@ -747,21 +747,24 @@ proc/anim(turf/location,atom/target,a_icon,a_icon_state as text,flick_anim as te
...
+			if(current.opacity)
+				return 0
```
ninjanomnom: FALSE

code/__HELPERS/unsorted.dm
```diff
@@ -747,21 +747,24 @@ proc/anim(turf/location,atom/target,a_icon,a_icon_state as text,flick_anim as te
...
+			for(var/thing in current)
+				var/atom/A = thing
+				if(A.opacity)
+					return 0
```
ninjanomnom: FALSE

code/__HELPERS/unsorted.dm
```diff
@@ -747,21 +747,24 @@ proc/anim(turf/location,atom/target,a_icon,a_icon_state as text,flick_anim as te
...
+			current = get_step_towards(current, target_turf)
+			steps++
 	return 1
```
ninjanomnom: TRUE

code/__HELPERS/unsorted.dm
```diff
@@ -747,21 +747,24 @@ proc/anim(turf/location,atom/target,a_icon,a_icon_state as text,flick_anim as te
...
-		if(current.opacity) return 0
-		for(var/atom/A in current)
-			if(A.opacity) return 0
+	var/steps = 1
+	if(current != target_turf)
 		current = get_step_towards(current, target_turf)
-		steps++
-
+		while(current != target_turf)
+			if(steps > length)
```
ninjanomnom: This may be faster done by checking distance outside the loop first

code/__HELPERS/unsorted.dm
```diff
@@ -747,21 +747,24 @@ proc/anim(turf/location,atom/target,a_icon,a_icon_state as text,flick_anim as te
...
+				return 0
+			if(current.opacity)
+				return 0
+			for(var/thing in current)
+				var/atom/A = thing
+				if(A.opacity)
+					return 0
+			current = get_step_towards(current, target_turf)
```
ninjanomnom: If you save the dir ahead of time you can just step in a direction instead of figuring out the direction every time with this proc

F-OS: Can you elaborate?

ninjanomnom: `get_step` is faster than `get_step_towards`

code/__HELPERS/unsorted.dm
```diff
@@ -747,21 +747,25 @@ proc/anim(turf/location,atom/target,a_icon,a_icon_state as text,flick_anim as te
...
 	var/turf/current = get_turf(source)
 	var/turf/target_turf = get_turf(target)
-	var/steps = 0
-
-	while(current != target_turf)
-		if(steps > length) return 0
-		if(current.opacity) return 0
-		for(var/atom/A in current)
-			if(A.opacity) return 0
+	var/steps = 1
```
ninjanomnom: You don't need this var anymore

ninjanomnom:
Now profile it again and make sure it works compared to the old one

F-OS:
Edited with new profile, tested to work with multiple situations and actually more accurate then the old version.

ninjanomnom:
I just realized I was being a dumbass with the get_step thing, not everything is necessarily a straight cardinal direction away from you

PR #246, opened 2018-08-29
Remove permasleep from tank treads.

Rohesie:
This is a small patch. I really dislike the method to stun targets here and the whole concept of sleep-inducing tank treads. This will have to be reworked later, but for now removing the bug will have to suffice.


F-OS:
what the fuck CM

Supermichael777:
should be knockdown and the treads should process to add knockdown to everything under them every tick

PR #247, opened 2018-08-30
get_area proc tweaked from recursion to iteration

Rohesie:
As per title.


Rohesie:
Add it to your own PR and I'll close this, @F-OS .

F-OS:
I was joking.

Issue #215, opened 2018-08-25
Auto-turret bullets blocked by barricades no matter the distance


Rohesie:
Kind of  a bug we inherited.

Also not entirely sure if we should keep the non-directional barricades. I'll revise that later.

Surrealaser:
This was resolved, wasn't it?

PR #248, opened 2018-08-30
Buff to the turrets' accuracy, damage and AP.

Rohesie:
Turrets are currently quite useless. Depending on how strong the change proves the dropship defenses may be tweaked separately, as they should not be easily boarded.

Also fixes https://github.com/ColonialMarines-Mirror/ColonialMarines-2018/issues/215

### DIFF
--- a/code/modules/cm_marines/dropship_equipment.dm
+++ b/code/modules/cm_marines/dropship_equipment.dm
@@ -19,41 +19,41 @@
 	var/ship_tag//used to associate the base to a dropship.
 	var/obj/structure/dropship_equipment/installed_equipment
 
-	Dispose()
-		if(installed_equipment)
-			cdel(installed_equipment)
-			installed_equipment = null
-		. = ..()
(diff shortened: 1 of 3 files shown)


Iatots:
This may be OOS for you but I'll put it here since you are rummaging in the sentry files:

CM turret pathing used to have an issue where the sentry's vision path was different than the bullet's path or something (basically, a beno around a corner would be visible to the turret, but the bullets would always path through the wall, making the turret waste them). They fixed it somehow, but then broke turrets: they would no longer fire through barricades and window frames (dense objects they could see through). They recently re-fixed this by switching from dense to opaque. The first instance is probably still in the codebase.

Rohesie:
Agreed with all, @Surrealaser .

I have not yet, @Iatots .

This was more of a quick patch since I'm working on something else at the moment. But, yes, those are issues to fix in the medium-to-long term.

Iatots:
WELP

Issue #222, opened 2018-08-26
Dropship Change Flight Type changing to wrong flight type


dabbster:
When clicking "Optimize Flight Plan" while FLYBY is on, it will result in the two flight types being mixed up, but both will result in Transport.
You will also receive conflicting messages, for example when you click Change Flight Type, the flight type in the dropship screen will show as FLYBY, but you will receive the message "You reset the flight plan to a transport mission between the Almayer and the planet.",  and when it changes to TRANSPORT you will get the message "You upload a flight plan for a low altitude flyby above the planet."

Not sure if this happens every time.

PR #249, opened 2018-08-30
Fix flyby/transport switching being inverted.

Rohesie:
Fixes https://github.com/ColonialMarines-Mirror/ColonialMarines-2018/issues/222

### DIFF
--- a/code/modules/shuttles/shuttle_console.dm
+++ b/code/modules/shuttles/shuttle_console.dm
@@ -181,7 +181,8 @@
 				to_chat(usr, "<span class='alert'>Hrm, that didn't work. Maybe try the one on the ship?</span>")
 				return
 			else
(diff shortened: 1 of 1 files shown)


code/modules/shuttles/shuttle_console.dm
```diff
@@ -180,13 +181,14 @@
 				to_chat(usr, "<span class='alert'>Hrm, that didn't work. Maybe try the one on the ship?</span>")
 				return
 			else
-				if(z == 1) shuttle.transit_gun_mission = 0 //remote launch always do transport flight.
+				if(z == 1) shuttle.transit_gun_mission = FALSE //remote launch always do transport flight.
```
ninjanomnom: missed one :P

PR #250, opened 2018-08-30
Now you can play Gloomy Sunday at round end.

Ghommie:
Exactly what the tin says, I see little to no reason to retain suicide restrictions to gun in mouth while throwing the EORG one out of the window on game over, or at least this is what I think. Thoughts are accepted but policy discussion should be avoided here. The maintainers have the final say.

TO DO: Citing this on either game completion and/or suicide verb input. Suggestions are accepted.
FUTURE TO DO: Unrestric the verb from humanity and make it available to xenos/simple animals/Robots.
FAR FUTURE MAYBE TO DO: Add/Port fancy self-termination methods.

### DIFF
--- a/code/game/verbs/suicide.dm
+++ b/code/game/verbs/suicide.dm
@@ -14,10 +14,8 @@
 
 	var/permitted = 0
 	var/list/allowed = list("Syndicate","traitor","Wizard","Head Revolutionary","Cultist","Changeling")
-	for(var/T in allowed)
-		if(mind.special_role == T || ticker.current_state == GAME_STATE_FINISHED)
-			permitted = 1
-			break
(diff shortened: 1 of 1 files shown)


code/game/verbs/suicide.dm
```diff
@@ -15,7 +15,7 @@
 	var/permitted = 0
 	var/list/allowed = list("Syndicate","traitor","Wizard","Head Revolutionary","Cultist","Changeling")
 	for(var/T in allowed)
-		if(mind.special_role == T)
+		if(mind.special_role == T || ticker.current_state == GAME_STATE_FINISHED)
 			permitted = 1
 			break
```
ninjanomnom: This loop is a bit horrifying. Check if `mind.special_role` is in the allowed list instead while you're here.

PR #251, opened 2018-08-31
Tricord now is slightly better at healing ( Has the same heal rate for all damages without that prob stuff) however it slightly slows you down.

F-OS:
:cl: 
Tricord will now confuse you if injected, however it heals better
:cl: 
Pros: May make people more protective of their combat medics.
Makes it so first-aid works better.
Seeing nerds stumble around drunk on the battlefield because they popped tricord is probably funny.
Cons: Not that enjoyable after the first few injections for people.
Might mess with the balance a bit.

### DIFF
--- a/code/modules/reagents/chemistry_reagents/medical.dm
+++ b/code/modules/reagents/chemistry_reagents/medical.dm
@@ -304,22 +304,26 @@
 	overdose = REAGENTS_OVERDOSE
 	overdose_critical = REAGENTS_OVERDOSE_CRITICAL
 
-	on_mob_life(mob/living/M, alien)
-		. = ..()
-		if(!.) return
-		if(M.stat == DEAD)
-			return
-		if(!alien)
-			if(M.getOxyLoss()) M.adjustOxyLoss(-REM)
-			if(M.getBruteLoss() && prob(80)) M.heal_limb_damage(REM, 0)
-			if(M.getFireLoss() && prob(80)) M.heal_limb_damage(0, REM)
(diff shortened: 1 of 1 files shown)


F-OS:
It seems that by screaming at travis i fixed it

huh

Iatots:
What if the confusion only happened at above 1 dose (15u)?

Rohesie:
Tricord is the basic healing tool of the marines untrained in medicine. It's supposed to be a weak and less-than-ideal general-purpose healing drug.
This will severely impact in the marines ability to heal and fight, but will buff backline healing.

I'm not really sold here and would like to see more discussion first.

GLACoding:
What Rohesie said, tricords supposed to be a injector you can pop in combat which can keep you going a while longer without having to stop and get real help from a medic.

Adding confusion means marines wont be able to push or retreat without getting screwed by RNG stumbling, which makes them dead weight to their squads until the tricord wears off (which for a 15 u injector takes quite a bit of time). This essentially means tricords only usable in the safety of an FOB where you probably don't need to move as much.

I haven't heard anyone complaining about the current level of healing from tricord, so I don't see the need to buff it either and people are already protective of their medics for their sweet trauma packs and splints.

F-OS:
Replaced the confusion with a slight slowdown.

Rohesie:
Tricord is a really good chem already. No need to buff it. And adding negative penalties doesn't seem to me that would bring many benefits, but would cause a lot of salt.

Make a new chem if you want instead. If it's balanced you can offer it as an alternative. Options are good as long as not game-breaking.

F-OS:
Good point.

PR #253, opened 2018-08-31
Reduces some of the roundstart pain for the PO,  they now have more powerloader training.

F-OS:
Pros: The PO now doesn't have to wait 5 minutes to move equipment
This may increase use of CAS
Cons: This may reduce RP with MTs and the CE.

### DIFF
--- a/code/datums/skills.dm
+++ b/code/datums/skills.dm
@@ -194,7 +194,7 @@
 /datum/skills/pilot
 	name = "Pilot Officer"
 	pilot = SKILL_PILOT_TRAINED
-	powerloader = SKILL_POWERLOADER_DABBLING
+	powerloader = SKILL_POWERLOADER_PRO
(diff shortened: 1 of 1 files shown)


Rohesie:
Moving stuff with a powerloader is neither too fun or game-breaking, but I wish we could find a more elegant solution to the problem than just maxing the PO's speed with it.

Do remember they can still use the powerloaders at dabbling, they are just a bit slower.

F-OS:
Do remember this isn't max speed, it's the same speed as the MTs, which makes sense because as the PO they would be loading and unloading dropships a ton. Also it's just good gameplay, as the slog of loading the DS pulls away from other enjoyable RP opportunities and gameplay.

Rohesie:
Don't have a counter for that. POs do have a reason to use the powerloader and this breaks nothing, nor impacts significantly balance-wise. Accepted.

PR #258, opened 2018-09-01
[READY-ish] Moves weapon stats out of the config and into the code

iskyp:
:cl: 
code: moved everything from updated_projectiles folder to projectiles because it was annoying to navigate
code: deleted ammo_datums.dm and moved everything from there into their own files to make finding what ammo type you're looking for easier.
code: replaced references to combat_defines.txt on bullets and guns with actual values
del: deleted most weapon related parameters in combat_defines.txt
/:cl:

why: no reason for these values to be hidden away, will make it easier for balancing as no one will assume that the server config is different, makes it easier to balance individual weapons instead of rebalancing everything that uses those config stats at once, makes the code for weapons and guns not a nightmare to read through

ready with probably a few stealth buffs and nerfs by accident
probably broken but it compiles and i did some minimal testing in game


code/modules/projectiles/magazines/pistols.dm
```diff
@@ -108,7 +108,7 @@
 /obj/item/ammo_magazine/pistol/holdout
 	name = "tiny pistol magazine (.22)"
 	desc = "A surprisingly small magazine, holding .22 bullets. No Kolibri, but it's getting there."
-	default_ammo = /datum/ammo/bullet/pistol/tiny
+	default_ammo = /datum/ammo/bullet/pistol
```
Rohesie: Is this a buff to the holdout pistol?

iskyp: nope pistol/tiny is actually the exact same as pistol from what i saw except in name

FantasticFwoosh:
It was probably done this way on part of a conspiracy to keep defacto CM closed repo _guests_ who weren't involved with the code directly in the dark deliberately, thanks for pulling them into the light.

iskyp:
ended up breaking everything on this branch when trying to resolve the conflicts just gonna pull on a new branch

PR #260, opened 2018-09-01
Prison Station Modification

Alterist:
A modification of Prison Station ready for live testing
Changes includes:
Moves DS2 to Civilian Hanger
Moves Xeno Spawns to the center of the map
Removes the first layer of podlocks from the western side
Adds tiny gardens that could be shoveled for dirt to LZ1 and Civilian Gardens


Rohesie:
Map bot is down. I'll review this when it's up. Feel free to poke me about it when it is.

Rohesie:
Since this completely alters the balance on Prison it will have to be playtested.

PR #262, opened 2018-09-01
Chemistry sanitizing

Rohesie:
Solutions courtesy of spookydonut.


Rohesie:
Clean the code for once. `src.occupant` is the same as `occupant` because `src` is implied if `occupant` is not a variable defined locally for the proc in question.

PR #264, opened 2018-09-01
Deletes flags_atom FPRINT, removes some redunant  flags_pass TABLEPASS etc.

Ghommie:
Title is explicit, this PR removes the FPRINT definition adn all its istances, and adjustes some TABLEPASS.

Unless I'm mistaken, FPRINT's scope was so that if an item didn't have the FPRINT flag, it wouldn't have retained any fingerprints on both add_fingerprint and add_hiddenprint (admin's version). By default all atoms had FPRINT on, and in no istance it was otherwise, so it's useless. A flag who does the opposite and null add_fingerprint would be better, but there is no reason to have one right now.

For the TABLEPASS, it's just removing some redunancy and something about a mutation that is currently unavailable due to the status of the code.

There is always the off chance I fouled up somewhere, reviews are (always) accepted.


code/__DEFINES/equipment.dm
```diff
@@ -19,7 +19,6 @@
 //flags_atom
 
 #define NOINTERACT				1		// You can't interact with it, at all. Useful when doing certain animations.
-#define FPRINT					2		// takes a fingerprint
 #define CONDUCT					4		// conducts electricity (metal etc.)
 #define ON_BORDER				8		// 'border object'. item has priority to check when entering or leaving
 #define NOBLOODY				16		// Don't want a blood overlay on this one.
```
Rohesie: Diminish each of these by a factor of 2 if you remove one in the middle. It's a bitflag after all.

PR #274, opened 2018-09-02
Move Aura Melee Damage Bonus

Surrealaser:
Per title.

Move Aura per Commands increases melee damage by +5 and +20% per rank of Leadership beyond the first.


FantasticFwoosh:
I re-iterate my standing position that i believe this change is unnessecary if you give a viable alternative, since anything from using *scream for a buff to adding in melee performance combat stimulants so that medbay keeps involved with the more active parts of the fight would be fine for combat damage without a aura eitherway when the marine needed it.

Rather than leave it open to stack as many buffs as they can gather together ontop of a move order to just reverse melee the :b:enos to death. Or suddenly forget to use primary fusilier tatics of ``shoot then retreat to a safe distance`` once the aura wears off and they die like idiots after rushing the xenos position.

Surrealaser:
@FantasticFwoosh In practice this works out to about +12 damage on a rifle/shotgun with a Bayonet.

Second, nothing's stopping medbay from making combat drugs.

Third, I like the idea of a thematic bayonet charge being make more viable.

Fourth, it's piggy backing off of an existing and thus far proven Orders system.

I wouldn't say this change is strictly necessary (the vast majority of changes aren't) but it does improve the general utility of Orders and is a meaningful addition to the mechanic which hopefully will see it get some further use.

Harcourt1:
Code seems quality now. Merge when ready.

code/_onclick/item_attack.dm
```diff
@@ -1,8 +1,6 @@
-
 // Called when the item is in the active hand, and clicked; alternately, there is an 'activate held object' verb or you can hit pagedown.
 /obj/item/proc/attack_self(mob/user)
 	return
-
```
ninjanomnom: Why are you removing all these lines for no reason?

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -121,7 +121,7 @@ Contains most of the procs that are called when a mob is attacked by something
 
 
 //Returns 1 if the attack hit, 0 if it missed.
-/mob/living/carbon/human/proc/attacked_by(var/obj/item/I, var/mob/living/user, var/def_zone)
+/mob/living/carbon/human/proc/attacked_by(var/obj/item/I, var/mob/living/user, var/def_zone, var/bonus_damage = 0)
```
ninjanomnom: `var/` is implicit in proc arguments

code/_onclick/item_attack.dm
```diff
@@ -1,74 +1,65 @@
...
 		power *= 2
-
 	if(user.mind && user.mind.cm_skills)
-		power = round(power * (1 + 0.3*user.mind.cm_skills.melee_weapons)) //30% bonus per melee level
-
+		bonus += round(power * 0.3 * max(0,user.mind.cm_skills.melee_weapons)) //30% bonus per melee level
+	if(ishuman(user))
+		var/mob/living/carbon/human/A = user
+		if(A.command_aura == "move") //Check for move aura; improves melee damage significantly.
+			bonus += 5 + round(power * max(0, A.command_aura_strength) * 0.20) // No cumulative multiplication with melee bonus. Move aura adds damage = 5 + 20% of base * Command level - 1, or +5 and +60% for a CO/XO.
```
Rohesie: Telebaton has `force = 50`.
Add 60% from old synth melee bonus and that gets +30
Add 5 + 60% and that gets +35.
Total damage: 125 brute per hit. Which bypasses armor.

Seems a bit excessive.

Surrealaser: Seems like that's more an issue with synths.

Also it's pretty much nothing compared to again, PB weapons, but I'll filter synths out though because they shouldn't be benefiting from orders as they're machines incapable of real emotion/inspiration.

Rohesie:
That is kind of a hack and it doesn't address the issue of stacking melee benefits. It's also inconsistent with Synths getting leadership benefits elsewhere, so you'd have to make that more ample if it was your intention.

I'm personally not convinced by the change. I'll wait for feedback.

Surrealaser:
> That is kind of a hack and it doesn't address the issue of stacking melee benefits. It's also inconsistent with Synths getting leadership benefits elsewhere, so you'd have to make that more ample if it was your intention.
> 
> I'm personally not convinced by the change. I'll wait for feedback.

It's a synth issue, not a general issue.

For who else would this even begin to approach being a problem? A CMP with the telescopic baton? Your standard grunt with a Bayo M37 is far deadlier in melee. Moreover, I've made the bonuses additive as opposed to multiplicative, to minimize problematic outliers.

If you want to insist on a global fix for Synths RE: orders I'll get around to it later.

ninjanomnom:
Rohesie is refering more to the code quality being a hack. A species proc like `can_receive_morale_bonuses()` would make sense for this so you can have a generic solution.

Surrealaser:
@ninjanomnom I'm aware; like I said, I'll submit a more comprehensive solution later.

ninjanomnom:
You just need to implement and use the generic solution here for now. That's a reasonable request. Maintainers are allowed to require quality improvements of related code within reason.

Rohesie:
Besides the code quality issue I feel this is not needed (move order is good as it is, requiring no buff) and makes balancing harder with having to take into account stacking melee benefits. Get a larger discussion with the playerbase to implement this. I'm against it myself, not feeling it adds but that it detracts.

Surrealaser:
> Besides the code quality issue I feel this is not needed (move order is good as it is, requiring no buff) and makes balancing harder with having to take into account stacking melee benefits. Get a larger discussion with the playerbase to implement this. I'm against it myself, not feeling it adds but that it detracts.

Of course you would, nevermind that there are only two melee buffs and they're already built in, and don't result in excessive damage whatsoever.

Rohesie:
Right, this is feature is denied for the above mentioned reasons.

PR #275, opened 2018-09-02
ADDED SHIT, REMOVED SHIT: ALMAYER EDITION

Alterist:
Adjusted the 2 Dropships to allow tank to enter if there's a rear hardpoint in place.
Added the Tank Fab to the tank hanger, also added a pair of welding googles.
Removed Spare Hardpoints parts in tank hanger
Put the scattered tools in tank hanger on racks, 'cause momma didn't raise no slob


francinum:
Since it seems mapbot has taken a holiday, can we get an image?

Alterist:
More space in the Dropships so that the Tank can fit inside no matter what

Neat and tidy Tank Bay, now with the Tank Fab

PR #278, opened 2018-09-02
Nukes smartgun bullets doing more damage with IFF off

### LINKED ISSUE #263: Smartgun and it's IFF meme.
Currently, having IFF off increases the damage and AP values of the SG for some stupid reason. 

Make it so the bonuses that are recived with IFF off are given to the SG when it has it's IFF on and then this stupid silly meme.

F-OS:
it was literally a 2 line change wtf cm devs

closes #263


Surrealaser:
Yes please.

Alterist:
smartgun penetration still increases with IFF off with your change. you might want to switch the armor_pen around, or make both the same

F-OS:
@Alterist It actually doesn't, IFF on penetration is mlow, IFF off it's low.

F-OS:
the actual fuck?

F-OS:
This is mainly so we have less FF from dumb SGers because LMAO IT DOES MORE DAMAGE KEK.

Borisvanmemes:
WHAT THE FUCK IT REALLY DOES DEAL MORE DAMAGE?!

code/modules/projectiles/updated_projectiles/ammo_datums.dm
```diff
@@ -724,17 +724,17 @@
...
 /datum/ammo/bullet/smartgun/lethal
 	flags_ammo_behavior = AMMO_BALLISTIC
 	icon_state 	= "bullet"
 
 /datum/ammo/bullet/smartgun/lethal/New()
 	..()
-	damage = config.lmed_hit_damage
-	penetration= config.low_armor_penetration
+	damage = config.low_hit_damage
+	penetration= config.mlow_armor_penetration
```
iskyp: why not literally just make it have the same stats, does turning IFF off somehow make the bullet weaker

F-OS: this is mainly a encouragement, a few SG bullets on a human will still down em so it works for HVH, but if you turn IFF off on xenos it will do less.

Harcourt1:
Friendly fire needs to be changed up anyways. I would merge this and i'm thinking about some code i was going to implement to lebensraum related to FF.

Rohesie:
The new changes to the smartgun have made it really powerful. Make these values the same in lethal and non-lethal, or at least don't make IFF stronger.

F-OS:
Done, I also lowered the penetration on IFF off to reduce FF

PR #281, opened 2018-09-03
The flamer now burns away vines.

F-OS:
Pros: 
More immersion.
More utility for the spec/sl.
Encourages people to be more protective of their SL/spec.

Cons: 
Straight marine buff.

### DIFF
--- a/code/modules/projectiles/updated_projectiles/guns/flamer.dm
+++ b/code/modules/projectiles/updated_projectiles/guns/flamer.dm
@@ -204,8 +204,6 @@
 			S.update_icon(1, 0)
 	
 	for(var/obj/structure/jungle/vines/V in T)
-		if(prob(25)) //Don't wanna spam this
-			V.visible_message("<span class='warning'>The vines burn away from the immense heat!</span>")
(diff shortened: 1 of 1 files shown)


Rohesie:
Machetes can already remove vines, this is not really a serious buff as fire blocks the path for quite some time. This is less efficient than the other method, so I see no issues balance-wise.

code/modules/projectiles/updated_projectiles/guns/flamer.dm
```diff
@@ -202,7 +202,12 @@
 		if (S.slayer > 0)
 			S.slayer -= 1
 			S.update_icon(1, 0)
-
+	
+	for(var/obj/structure/jungle/vines/V in T)
+		if(prob(25)) //Don't wanna spam this
+			V.visible_message("<span class='warning'>The vines burn away from the immense heat!</span>")
```
ninjanomnom: Why do we need a message?

F-OS: Eh good point.

PR #285, opened 2018-09-04
Queen locator shows Hive Status when clicked

JJRcop:
this allows most xenos to not need to use the Alien tab

### DIFF
--- a/code/_onclick/hud/screen_objects.dm
+++ b/code/_onclick/hud/screen_objects.dm
@@ -668,8 +668,16 @@
 /obj/screen/queen_locator
 	icon = 'icons/mob/screen1_alien.dmi'
 	icon_state = "trackoff"
-	name = "queen locator"
+	name = "queen locator (click for hive status)"
 	screen_loc = ui_queen_locator
+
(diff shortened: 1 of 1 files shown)


code/_onclick/hud/screen_objects.dm
```diff
@@ -668,9 +668,17 @@
 /obj/screen/queen_locator
 	icon = 'icons/mob/screen1_alien.dmi'
 	icon_state = "trackoff"
-	name = "queen locator"
+	name = "queen locator (click for hive status)"
 	screen_loc = ui_queen_locator
 
+/obj/screen/queen_locator/clicked(var/mob/user)
+	if (..())
+		return 1
```
Rohesie: `return TRUE`

code/_onclick/hud/screen_objects.dm
```diff
@@ -668,9 +668,17 @@
...
+	if (isXeno(user))
+		var/mob/living/carbon/Xenomorph/X = user
+		X.hive_status()
+	return 1
```
Rohesie: `return TRUE`

PR #286, opened 2018-09-04
Moves weapon stats out of the config and into the code

iskyp:
:cl:
code: deleted ammo_datums.dm and moved everything from there into their own files to make finding what ammo type you're looking for easier.
code: replaced references to combat_defines.txt on bullets and guns with actual values
del: removes combat_defines.txt
/:cl:

why: no reason for these values to be hidden away, will make it easier for balancing as no one will assume that the server config is different from their local config, makes it easier to balance individual weapons instead of rebalancing everything that uses those config stats at once, makes the code for weapons and guns not a nightmare to read through

ready with probably a few stealth buffs and nerfs by accident
probably broken but it compiles and i did some minimal testing in game
if i missed anything tell me
also testmerging this is probably a good idea because it changes alot and i'd just like to make sure it works on a full round

it's just #258 but hopefully not going to break if i have to rebase because any changes on ammo = merge conflicts


iskyp:
lazy and don't care to finish this

PR #287, opened 2018-09-05
Updated Med HUD So It Ignores Ghostless Bodies

Surrealaser:
Per title.


Surrealaser:
Fixed here: https://github.com/ColonialMarines-Mirror/ColonialMarines-2018/pull/300

PR #288, opened 2018-09-05
(Still unimplemented) Benos' Salvage Plasma ability.

Ghommie:
Nothing too special, coded something for the critters: The ability of salvaging/absorbing plasma from deceased sisters with an efficiency rate of 50% and a delay (both placeholders). not normally achievable so far as no caste has it (But locally tested it on some, works fine).  
The Idea is for benos to have some sort of counterpart of marine's corpse looting and beno body selling, as right now a dead comrade is nothing more than trash on the floor for them. There will be the need of balancing some stats here and there, unhide the plasma bar of dead benos, and others.

I'm fairly sure the code is a bit spaghetti, I was "heavily inspired" from the transfer_plasma lines just above. Reviews very welcomed.


Rohesie:
I like the concept.

GLACoding:
I was going to code in spitters being able to acid dissolve dead sisters, but this is a much better idea.

Ghommie:
A little tweak added, the bugs' plasma slowly starts to decay after their death. No forever fresh juice.

Ghommie:
Ok, I updated this one after a while. 
- "Robotic" xenos can't transfer or absorb plasma to/from "organic" ones.
- Handle_decay() should now deal with the slow plasma depletion of dead xenos (minus xenoborgs/larvas).
- Also moved some caste proc overrides from Xenomorph/life.dm to their respective caste files.

code/modules/mob/living/carbon/xenomorph/Abilities.dm
```diff
@@ -416,6 +416,27 @@
 	transfer_delay = 5
 	max_range = 7
 
+/datum/action/xeno_action/activable/salvage_plasma
+	name = "Salvage Plasma"
+	action_icon_state = "salvage_plasma"
+	ability_name = "salvage plasma"
+	var/plasma_salvage_amount = 40
+	var/salvage_delay = 50
```
Rohesie:
```suggestion
	var/salvage_delay = 5 SECONDS
```

code/modules/mob/living/carbon/xenomorph/Abilities.dm
```diff
@@ -416,6 +416,27 @@
...
+	var/mob/living/carbon/Xenomorph/X = owner
+
+	if(owner.action_busy)
+		return
+
+	X.xeno_salvage_plasma(A, plasma_salvage_amount, salvage_delay, max_range)
+
+datum/action/xeno_action/activable/salvage_plasma/hivelord
+	plasma_salvage_amount = 80
+	salvage_delay = 30
```
Rohesie:
```suggestion
	salvage_delay = 3 SECONDS
```

code/modules/mob/living/carbon/xenomorph/Castes/Predalien.dm
```diff
@@ -1,4 +1,4 @@
-/mob/living/carbon/Xenomorph/Predalien
+/mob/living/carbon/Xenomorph/Predalien/
```
Rohesie: Why this change?

Rohesie (changes requested):

Rohesie:
There is a xenomorph rework in process, turning castes into datums. You might want to speak to @spookydonut about it so your work doesn't overlap.

code/modules/mob/living/carbon/xenomorph/Castes/Predalien.dm
Ghommie: A leftover from copypasting the path for procs.

code/modules/mob/living/carbon/xenomorph/Powers.dm
```diff
@@ -984,10 +989,67 @@
...
 	target.gain_plasma(amount)
-	to_chat(target, "<span class='xenowarning'>\The [src] has transfered [amount] plasma to you. You now have [target.plasma_stored].</span>")
-	to_chat(src, "<span class='xenowarning'>You have transferred [amount] plasma to \the [target]. You now have [plasma_stored].</span>")
+	to_chat(target, "<span class='xenowarning'>\The [src] has transfered [amount] units of [is_robotic ? "power" : "plasma"] to you. You now have [target.plasma_stored]/[target.plasma_max].</span>")
+	to_chat(src, "<span class='xenowarning'>You have transferred [amount] units of [is_robotic ? "power" : "plasma"] to [target]. You now have [plasma_stored]/[plasma_max].</span>")
 	playsound(src, "alien_drool", 25)
 
+/mob/living/carbon/Xenomorph/proc/xeno_salvage_plasma(atom/A, amount, salvage_delay, max_range)
+
+	if(!istype(A, /mob/living/carbon/Xenomorph) || !check_state())
```
Rohesie:
```suggestion
	if(!isXeno(A) || !check_state())
```

Rohesie (changes requested):

Rohesie:
If there are enough xenos, this will add a bit of processing, as it will have to do extra process all the dead corpses. We were thinking of turning the living xenos into a list in order to handle life more efficiently. We could turn this into a subsystem which is called with much less priority.
There should be a check to return, though, or to send it to a different list with no more processing, after the plasma has ran dry. Or to gib them then.

PR #291, opened 2018-09-05
QoL: Auto-Drop Items in Offhand When Wielding

Surrealaser:
Per title. QoL change.


Surrealaser:
@iskyp What the problem is?

mattroks101:
It should try to equip the item first, then drop it.

Surrealaser:
@mattroks101 I considered it during my first draft; I'm not sure I'm into the idea of a panic wield involving an attempted careful placement prior to dropping, but it seems as though there are several people now who want this, so I'll add it.

Surrealaser:
@factoryman942 Done.

PR #294, opened 2018-09-07
New smoke grenades that actually makes the passerbys (almost) invisible to benos. As seen on TV.

Ghommie:
On behalf of the masses' complaints against the outdated HSDP grenades, I spent this day coding and testing and coding a the new brand SCDP (Smoke Screen ""Dual Purpose"") grenade, which is guaranteed to shielding against thermal vision to all mobs minus ghosts. Code isn't of the prettiest should be devoid of (hopefully) bugs and runtimes and shouldn't give off permanent cloaking on mundane occasions. 
Currently only adminspawnable for the delicate alpha reason, if anyone wants to test it, try going out the norms, tele in and out, admindelete the smoke, and report any problem or review.
TODO: Sprite the new grenade.

### DIFF
--- a/code/game/objects/effects/effect_system/smoke.dm
+++ b/code/game/objects/effects/effect_system/smoke.dm
@@ -119,6 +119,88 @@
 				M.coughedtime = 0
 
 /////////////////////////////////////////////
+// Cloak Smoke (SEE_MOBS Hur Dur)
+////////////////////////////////////////////
+/obj/effect/particle_effect/smoke/tactical
+
+/obj/effect/particle_effect/smoke/tactical/New(loc, oldamount)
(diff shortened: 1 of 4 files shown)


Iatots:
Does this smoke still block vision? Remember that benos can easily see the cloak outline in the pure black of unseen tiles.

Ghommie:
Improved the SCDP code a bit, now it is sturdier against most (if not all) game breaking exploits. The update_cloak() takes care of it so that if you somehow still end up cloaked outside the smoke area, it will force the uncloaking proc after 2 seconds or so. Still would like a review here and there to clean up the code. Sprite(s) coming up soon

PR #295, opened 2018-09-07
The pyro spec no longer gets fucked by angry tackle-spam xenos dragging him into his own fire.

F-OS:
Title.

Pros: Removes a infuriating meta

Cons: Fucks with the general balance and skill floor for pyro specs.

### DIFF
--- a/code/modules/clothing/head/helmet.dm
+++ b/code/modules/clothing/head/helmet.dm
@@ -324,7 +324,7 @@
 /obj/item/clothing/head/helmet/marine/pyro
 	name = "\improper M35 helmet"
 	icon_state = "pyro_helmet"
-	desc = "A helmet designed for USCM Pyrotechnicians."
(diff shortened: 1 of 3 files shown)


Surrealaser:
@F-OS If this makes him immune to fire, this is long overdue, and I approve.

Supermichael777:
So even more reason for the brainlet pyro main to stand in the back and torch his own squad

Rohesie:
We'll have to test this and tweak how much burn damage they get. It's fine that they don't get set on fire, but we don't want them to be 100% immune to fire through kelo+derm+tram, for example.

Surrealaser:
@Rohesie They should be basically immune to fire; nothing wrong with that.

Rohesie:
They have the best armor in the game second only to B18. Total fire immunity might be a bit too much.

PR #300, opened 2018-09-09
Fix for Updated Med HUD So It Ignores Ghostless Bodies

Surrealaser:
Fix for: https://github.com/ColonialMarines-Mirror/ColonialMarines-2018/pull/287

### DIFF
--- a/code/datums/mob_hud.dm
+++ b/code/datums/mob_hud.dm
@@ -271,9 +271,12 @@ var/datum/mob_hud/huds = list(
 		if(world.time - timeofdeath > revive_grace_period || undefibbable)
 			revive_enabled = 0
 		else
-			var/mob/dead/observer/G = get_ghost()
-			if((istype(G) && !client) || !suiciding || !head || !head.is_usable() || !heart || heart.is_broken() || !has_brain() || chestburst || (HUSK in mutations) || !mind)
+			if(suiciding || !head || !head.is_usable() || !heart || heart.is_broken() || !has_brain() || chestburst || (HUSK in mutations) || !mind)
 				revive_enabled = 0
+			else if(!client)
+				var/mob/dead/observer/G = get_ghost()
+				if(!istype(G))
+					revive_enabled = 0
 
 		var/holder2_set = 0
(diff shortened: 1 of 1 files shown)


Kingtrin:
The first version of this code broke inexplicably and had us very confused in dmca-coding. It is based off of badly written code that shouldn't work to begin with, but it does for reasons beyond the comprehension of science. Beware ye all who enter here.

Supermichael777:
Okay so this is a DISCONNECT detector, It is Flawed as a use for that, and I suspect that the Code this draws its example from is terminally flawed, as code that runs before the code this draws from works and would EXCLUDE THE CHECK THIS IS BASED ON FROM RUNNING.

Relevent checks in order

```
if(istype(G))
 ....
else if(!H.client)
    return

if(!H.client)
    mesage
*if(H.mind && !H.client) //Let's call up the correct ghost! Also, bodies with clients only, thank you.
    G = H.get_ghost()
    if(istype(G))
          mesage
```


From this, including the note i conclude that a clientless OR mindless body will never have a ghost, and will also never be revivable, and that the check marked with a * contains an error.  I suspect the original check errored true because BYOND didn't like the way it was formatted.

NOTE HOWEVER THAT BODIES WITHOUT GHOSTS MAY BE REVIVABLE AS GHOSTS ARE CREATED BY A PLAYER CONTROLLED BUTTON AFTER DEATH.  PLAYERS THAT DIE DO NOT IMMEDIATELY HAVE GHOSTS.  HAVING A GHOST IS NOT A RELIABLE WAY TO CHECK THE STATE OF A PLAYER.

THE GHOST BASED CONTROL STRUCTURE ONLY GIVES WARNINGS TO GHOSTS. GET GHOST FILTERS OUT PLAYERS THAT CANNOT RE-ENTER BUT IT ALSO FILTERS OUT PLAYERS ALREADY IN THE BODY.  IT SHOULD ONLY BE USED FOR POSITIVE DETECTION OF PLAYERS WHO:

1. Have ghosted
2. Can re-enter the body

Issue #118, opened 2018-08-18
underwear selection doesnt work


chesse20:
i think it doesn't save so if you dont want your female character to wear boxers you have to change your underwear everytime

PR #303, opened 2018-09-11
Fixes women getting mens underwear in their saves

Supermichael777:
Fixes: #118
Desktop git is an ass to work with sometimes, especially with DMs need to run its shit in an identically named file


FantasticFwoosh:
I thought all the lingerie was universal by design to the settings, its a nice customisation QOL change anyway.

PR #307, opened 2018-09-14
Now you can quick equip your weaponary and ammo in holsters and scabbards.

Ghommie:
Now you can just press E to equip your weapons and ammo to an appropriate holster/scabbard.

As a gamer, I don't really like the lack of quick storaging for all containers, but on the other hand that would null the little comfort this PR adds to holsters and the distasteful large holsters. I don't enjoy this code too much, but it works.

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -455,7 +455,25 @@ cases. Override_icon_state should be a list.*/
 			if(WEAR_IN_BACK)
 				if (H.back && istype(H.back, /obj/item/storage/backpack))
 					var/obj/item/storage/backpack/B = H.back
-					if(B.can_be_inserted(src))
+					if(src.w_class <= B.max_w_class)
+						if(B.can_be_inserted(src))
+							return 1
+			if(WEAR_IN_B_HOLSTER)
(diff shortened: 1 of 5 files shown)


code/game/objects/items.dm
```diff
@@ -455,7 +455,25 @@ cases. Override_icon_state should be a list.*/
 			if(WEAR_IN_BACK)
 				if (H.back && istype(H.back, /obj/item/storage/backpack))
 					var/obj/item/storage/backpack/B = H.back
-					if(B.can_be_inserted(src))
+					if(src.w_class <= B.max_w_class)
+						if(B.can_be_inserted(src))
+							return 1
```
Rohesie: This is a TRUE/FALSE question. Mind altering those returns?

code/modules/mob/living/carbon/human/inventory.dm
```diff
@@ -77,6 +77,12 @@
 			return 1
 		if(WEAR_IN_ACCESSORY)
 			return 1
+		if(WEAR_IN_HOLSTER)
+			return 1
```
Rohesie: Same in this file. This is a TRUE value.

PR #321, opened 2018-09-16
Puts Supply Pads on Layer 2, fixing z-fight issues.

### LINKED ISSUE #317: Supply Pads are Layer 3
This can cause it to fight for superiority over what you're actually trying to drop down. They need to be Layer 2.

francinum:
fixes #317

### DIFF
--- a/code/modules/cm_marines/overwatch.dm
+++ b/code/modules/cm_marines/overwatch.dm
@@ -856,7 +856,7 @@
 	anchored = 1
 	density = 0
 	unacidable = 1
-	layer = 2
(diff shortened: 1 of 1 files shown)


code/modules/cm_marines/overwatch.dm
```diff
@@ -856,6 +856,7 @@
 	anchored = 1
 	density = 0
 	unacidable = 1
+	layer = 2
```
Ghommie: layer = ABOVE_TURF_LAYER

PR #324, opened 2018-09-17
 StaggerStun Proc + Shotgun Munition Rework

Surrealaser:
-StaggerStun Proc reworked to differentiate between hard CC and soft CC, featuring different requisite variables for each. Soft CC (slowdown and ability lock) now works on all castes by default.

Further, targets that are already stunned/knocked down can't be stunned/knocked down by the Staggerstun proc.

-Slug support aspects improved; slow and stagger soft CC effects increased to 8 each up from 2.

-After playtesting flechettes determined as needing further improvement, thus:
1. Flechette falloff reduced by 50%; -0.5 damage per tile down from -1 per tile.
2. Damage increased for each flechette projectile to 40 up from 35.
3. Max range for secondary flechettes increased to 10 (not sure why it was reduced to 5).
4. Flechette scatter chance for secondary flechettes reduced to 40% from 80%.

### DIFF
--- a/code/modules/projectiles/updated_projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/updated_projectiles/ammo_datums.dm
@@ -595,7 +595,7 @@
 	accuracy_var_low = config.med_proj_variance
 	accuracy_var_high = config.med_proj_variance
 	max_range = config.short_shell_range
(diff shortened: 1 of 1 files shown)


Rohesie:
Considerable change, we'll have to playtest it.

PR #325, opened 2018-09-17
Spec armor no longer does negative damage

F-OS:
what the fuck CM

### DIFF
--- a/code/modules/clothing/head/helmet.dm
+++ b/code/modules/clothing/head/helmet.dm
@@ -301,7 +301,7 @@
 /obj/item/clothing/head/helmet/marine/specialist
 	name = "\improper B18 helmet"
 	desc = "The B18 Helmet that goes along with the B18 Defensive Armor. It's heavy, reinforced, and protects more of the face."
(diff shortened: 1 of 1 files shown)


Rohesie:
This is not a % reduction. If you want to understand how armor works check these:
https://github.com/ColonialMarines-Mirror/ColonialMarines-2018/blob/5f44f28548cdadaa9748fd8291b49dd30198ccd8/code/modules/mob/living/carbon/human/human_defense.dm
(specifically the `getarmor_organ` proc)
https://github.com/ColonialMarines-Mirror/ColonialMarines-2018/blob/master/code/modules/projectiles/updated_projectiles/projectile.dm
(the `human/bullet_act` proc)

F-OS:
I know, i was just joking about the 105%, and considering that only admin spawned deathsquad helmets have more bullet resistance then this, i think it can be safely tuned down.

Rohesie:
Sorry, I kind of don't see the need for this change. It's not as if we want the B18 to be a little more vulnerable to bullets in the head, even if only a tiny bit.

F-OS:
Good point.

PR #327, opened 2018-09-17
Vines and Fire Sprite Update

Alterist:
Changes out the spacevines.dmi and onfire.dmi to tg versions.
Special thanks to Iatots
untitled-1
untitled-2
untitled-3
untitled-5


Ghommie:
Some of the space vines from the second example picture of the lodge seem to missing, is that from some non commited map change?

Also the fire would need some darker/red outlines and bits, as it seems to quite stressing to distinguish the mob from a standard flamed turf.

Alterist:
untitled-1
That was an old pic, here's one for proper comparison

Rohesie:
I'll let it sit a little and then approve them if I see no issues raised against it.

PR #328, opened 2018-09-17
Fixes trivial things reported on CM.

Ghommie:
This quite unatomized PR fixes the easiest, minor stuff previously reported on CM on the past month.
If I missed something feel free to shout out.


Harcourt1:
At least we get shit done here at /tg/. Code seems good.

PR #329, opened 2018-09-18
If you're jobbanned you can't join a xeno response team.

F-OS:
Pros: Consistency 

Cons: Literally none.

### DIFF
--- a/code/datums/emergency_calls/xenos.dm
+++ b/code/datums/emergency_calls/xenos.dm
@@ -21,8 +21,10 @@
 /datum/emergency_call/xenos/create_member(datum/mind/M)
 	var/turf/spawn_loc = get_spawn_point()
 	var/mob/original = M.current
-
-	if(!istype(spawn_loc)) return //Didn't find a useable spawn point.
+	if(jobban_isbanned(original, "Alien"))
+		return
+	if(!istype(spawn_loc))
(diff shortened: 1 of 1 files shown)


code/datums/emergency_calls/xenos.dm
```diff
@@ -21,8 +21,10 @@
 /datum/emergency_call/xenos/create_member(datum/mind/M)
 	var/turf/spawn_loc = get_spawn_point()
 	var/mob/original = M.current
-
-	if(!istype(spawn_loc)) return //Didn't find a useable spawn point.
+	if(jobban_isbanned(original, "Alien"))
+		return
```
Rohesie: While this does prevent xeno job-baned players from playing a XRT, it also eats up the slot if one was selected.
A way around that is to `return FALSE` here (for clarity) and then alter the logic of the loop that calls this proc, which you'll find in `/datum/emergency_call/proc/activate(announce = TRUE)`, in `emergency_call.dm`

F-OS: Ah, got it, thanks again!

Rohesie:
Do you plan on continuing this?

F-OS:
Sorry, forgot about it, will fix soon :tm:

F-OS:
Press :f:

PR #330, opened 2018-09-18
Fixes shotguns hitting twice in melee when without ammo and on harm intent.

Rohesie:
Small bug in PB code.


Rohesie:
Should be ready for review, @ninjanomnom .

PR #331, opened 2018-09-18
Replaces Anesthetic Injectors with Surgical Injectors, now with 100% less black screen!

F-OS:
Pros:
1. Removes a tool that's normally only used for grief.
2. Makes field surgery more tolerable
Cons:
1. More access to pain killers.
2. Reduces the downsides of field surgery.

### DIFF
--- a/code/game/machinery/vending/marine_vending.dm
+++ b/code/game/machinery/vending/marine_vending.dm
@@ -293,9 +293,10 @@
 					/obj/item/stack/medical/ointment = 8,
 					/obj/item/stack/medical/splint = 2,
 					/obj/item/device/healthanalyzer = 3,
+					/obj/item/reagent_container/hypospray/autoinjector/surgical = 3
 					/obj/item/bodybag/cryobag = 2)
 
-	contraband = list(/obj/item/reagent_container/hypospray/autoinjector/chloralhydrate =3)
+	contraband = list(/obj/item/reagent_container/pill/happy = 4)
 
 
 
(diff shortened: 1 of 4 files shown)


Rohesie:
This is not very practical because both paracetamol and spaceacillin have a very low metabolization rates, while oxy has very high ones. Sure, you can use 6 injectors before you run into trouble, but still, after that you'll leave the patient with chems that won't go away in a long, long time. And unable to use tramadol at risk of getting toxins.

There is also the fact that painmed surgery is not very effective, and we haven't yet changed that.

F-OS:
Alright, lemme adjust it a bit, thanks!

Kingtrin:
Having spaceacillin remain in the body after field surgery is ideal actually do the high infection rates. The main downside to field surgery is needing to do mask+anesthetic tank every time so having a quick alternative when shit is hitting the fan would be nice. If a step fails you just recheck the bones and keep going.

Rohesie:
Current implementation is both a combat drug and makes the oxycodone injectors obsolete.
Ironically, the solution to that would be re-adding the sleeping agent chem that was removed from them initially, even if in lower doses.

I'd remove the spaceacilin. You can carry a pill bottle with 16 of those, or two if you feel one is not enough. Give options to the doctors, don't force them into a single one.
Also the paracetamol. It means you can't operate on someone with tramadol on their system.

F-OS:
Actually, lemme code in a new chem that'll make these useful without them becoming a combat stim.

Rohesie:
Make something similar to oxycodone: very quick metabolization, low OD threshold, reduces flinching. You can go with it selectively numbing parts of the reflex arc for an explanation on why it doesn't also reduce pain. If you feel creative there's side-effects you can concoct based on that.

F-OS:
a lotta good points got raised

PR #332, opened 2018-09-18
You can now order a heavy pulse rifle for 25 points as the SL.

F-OS:
Pros:
1. The HPR almost never sees use due to it's price and i feel that it's a interesting tool that hasn't been explored.
2. Increases the value of the SL.
Cons:
1. Marine buff.
2. Gives a incentive for SLs to be more aggressive, which may cause them to be killed.


Rohesie:
Despite how much noise it makes, the HPR is not really a great gun. It's more of a special snowflake mediocre gun.
I've seen people going SL in the past in order to get preferential treatment in requisitions (getting HPR as well) instead of leading.
Balance-wise I'm not to worried. I'd never pick this as SL myself. My concern is on the incentives of giving rare combat tools to the ones who should be leading, not playing to be the main protagonists in their robust story.

Discuss.

Rohesie:
Already handled in #372

PR #334, opened 2018-09-18
Adds a marine vendor for PFCs

F-OS:
Pros:
1. Reduces roundstart DMV simulator 2018
2. Allows for more varied gameplay.

Cons:
1. Slight buff.


Rohesie:
Free mag harnesses might be an issue. The rest is fine, IMO.

F-OS:
Fixed as per feedback.

Rohesie:
Odd selection. Why these? These are not really the most common attachments by far.

F-OS:
Mainly because i feel these attachments are able to be given out without balance issues.

Rohesie:
Recoil compensator, angled grip, laser sight, gyro, bipod, any of the stocks and the mini-flamer all fit the same bill. Req never runs out of any of them.

Rohesie:
In the clothes, what?

F-OS:
It was the only thing that worked quickly and elegantly.

PR #336, opened 2018-09-19
Smartgun Improvement

Surrealaser:
Ammo Capacity: 500 pack and 100 belt. Up from 250 and 50.

 AP 30 up from 10 so it can actually deal with decent armour, instead of plinking it for barely any damage.

Falloff damage reduction per tile -0.5 down from -1 so it doesn't lose 10-20% of its already low base damage for being backline as it should be.

Can now take the following attachments:
* Extended Barrel
* Flashlight (Starts with this by default)

Automatic reload system actually works on the move as god intended, but always does so at the slowest speed because skills don't apply.

Removed the thermals overlay for the SG visor since no one liked it, and it hurt peoples' eyes Virtua Boy style.

### DIFF
--- a/code/modules/cm_marines/equipment/weapons.dm
+++ b/code/modules/cm_marines/equipment/weapons.dm
@@ -74,11 +74,17 @@
 				reload(user, mygun)
 			else
 				to_chat(user, "Your reloading was interrupted!")
+				playsound(src,'sound/machines/buzz-two.ogg', 25, 1)
 				reloading = 0
 				return
 		else
-			sleep(reload_duration)
-			reload(user, mygun)
+			if(autoload_check(user, reload_duration, mygun, src))
+				reload(user, mygun, TRUE)
+			else
+				to_chat(user, "The automated reload process was interrupted!")
+				playsound(src,'sound/machines/buzz-two.ogg', 25, 1)
+				reloading = 0
+				return
 		return 1
 
 	attackby(var/obj/item/A as obj, mob/user as mob)
@@ -99,7 +105,7 @@
 			if(pcell)
 				to_chat(user, "A small gauge in the corner reads: Ammo: [rounds_remaining] / [rounds_max].")
 
-/obj/item/smartgun_powerpack/proc/reload(mob/user, obj/item/weapon/gun/smartgun/mygun)
+/obj/item/smartgun_powerpack/proc/reload(mob/user, obj/item/weapon/gun/smartgun/mygun, automatic = FALSE)
 
 		pcell.charge -= 50
 		if(!mygun.current_mag)
@@ -111,11 +117,29 @@
 		mygun.current_mag.current_rounds += rounds_to_reload
 		rounds_remaining -= rounds_to_reload
 
-		to_chat(user, "You finish loading [rounds_to_reload] shells into the M56 Smartgun. Ready to rumble!")
+		if(!automatic)	to_chat(user, "You finish loading [rounds_to_reload] shells into the M56 Smartgun. Ready to rumble!")
+		else	to_chat(user, "The powerpack servos finish loading [rounds_to_reload] shells into the M56 Smartgun. Ready to rumble!")
 		playsound(user, 'sound/weapons/unload.ogg', 25, 1)
 
 		reloading = 0
 		return 1
+
+/obj/item/smartgun_powerpack/proc/autoload_check(mob/user, delay, obj/item/weapon/gun/smartgun/mygun, obj/item/smartgun_powerpack/powerpack, numticks = 5)
+	if(!istype(user) || delay <= 0) return FALSE
+
+	var/mob/living/carbon/human/L
+	if(istype(user, /mob/living/carbon/human)) L = user
+
+	var/delayfraction = round(delay/numticks)
+	. = TRUE
+	for(var/i = 0 to numticks)
+		sleep(delayfraction)
+		if(!user)
+			. = FALSE
+			break
+		if(!(L.back == powerpack) || !(L.s_store == mygun) && !(user.get_active_hand() == mygun)) //power pack and gun aren't where they should be.
+			. = FALSE
+			break
 
 /obj/item/smartgun_powerpack/snow
 	icon_state = "s_powerpack"
--- a/code/modules/cm_marines/equipment/weapons.dm
+++ b/code/modules/cm_marines/equipment/weapons.dm
@@ -122,7 +122,7 @@
 		playsound(user, 'sound/weapons/unload.ogg', 25, 1)
 
 		reloading = FALSE
-		return 1
+		return TRUE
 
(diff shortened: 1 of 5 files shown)


Surrealaser:
@Rohesie BTW, the issue with the autoloader is fixed; it now checks to see if you have the power pack and gun on your person; if either is missing, the auto-reload will be interrupted.

code/modules/cm_marines/equipment/weapons.dm
```diff
@@ -103,7 +103,43 @@
...
+	if(istype(user, /mob/living/carbon/human)) L = user
+
+	var/delayfraction = round(delay/numticks)
+	. = TRUE
+	for(var/i = 0 to numticks)
+		sleep(delayfraction)
+		if(!user)
+			. = FALSE
+			break
+		if(!(L.back == powerpack) || !(L.s_store == mygun) && !(user.get_active_hand() == mygun)) //power pack and gun aren't where they should be.
```
Rohesie: This will cause the user to keep reloading without a powerpack as long as the gun remains in the hand.
To accomplish what you want perhaps checking both hands for the gun might be ideal.

code/modules/projectiles/updated_projectiles/guns/specialist.dm
```diff
@@ -203,22 +203,26 @@
...
-	var/shells_fired_max = 20 //Smartgun only; once you fire # of shells, it will attempt to reload automatically. If you start the reload, the counter resets.
+	var/shells_fired_max = 50 //Smartgun only; once you fire # of shells, it will attempt to reload automatically. If you start the reload, the counter resets.
 	var/shells_fired_now = 0 //The actual counter used. shells_fired_max is what it is compared to.
 	var/restriction_toggled = 1 //Begin with the safety on.
 	gun_skill_category = GUN_SKILL_SMARTGUN
 	attachable_allowed = list(
+						/obj/item/attachable/extended_barrel,
 						/obj/item/attachable/heavy_barrel,
+						/obj/item/attachable/reddot,
+						/obj/item/attachable/flashlight,
```
Rohesie: Some of these attachments are not much a case of trade-off options, but must-have instead. And smartgunners have night vision.
Also adding attachments because they look cool is nice and all, but try to get the sprite itself improved if that's what you want.

code/modules/cm_marines/equipment/weapons.dm
Surrealaser: Yes, you know this because I told you. :p

code/modules/projectiles/updated_projectiles/guns/specialist.dm
Rohesie: More like yet another buff, and not a trade off option.

Surrealaser: It's fine. How many times have we done this, where you're worried about something being 'totally broken', I say its not, and it gets playtested and turns out to be fine?

code/modules/cm_marines/equipment/weapons.dm
```diff
@@ -61,30 +62,29 @@
...
-			return 1
+		if(!automatic)
+			if(user.mind && user.mind.cm_skills && user.mind.cm_skills.smartgun>0)
+				reload_duration = max(reload_duration - 10*user.mind.cm_skills.smartgun,30)
+			if(do_after(user,reload_duration, TRUE, 5, BUSY_ICON_FRIENDLY))
+				reload(user, mygun)
+			else
+				to_chat(user, "Your reloading was interrupted!")
+				playsound(src,'sound/machines/buzz-two.ogg', 25, 1)
+				reloading = 0
```
Rohesie: FALSE

code/modules/cm_marines/equipment/weapons.dm
```diff
@@ -61,30 +62,29 @@
...
 		else
-			to_chat(user, "Your reloading was interrupted!")
-			reloading = 0
-			return
+			if(autoload_check(user, reload_duration, mygun, src))
+				reload(user, mygun, TRUE)
+			else
+				to_chat(user, "The automated reload process was interrupted!")
+				playsound(src,'sound/machines/buzz-two.ogg', 25, 1)
+				reloading = 0
```
Rohesie: FALSE

code/modules/cm_marines/equipment/weapons.dm
```diff
@@ -103,7 +103,43 @@
...
+		var/rounds_to_reload = min(rounds_remaining, (mygun.current_mag.max_rounds - mygun.current_mag.current_rounds)) //Get the smaller value.
+
+		mygun.current_mag.current_rounds += rounds_to_reload
+		rounds_remaining -= rounds_to_reload
+
+		if(!automatic)	to_chat(user, "You finish loading [rounds_to_reload] shells into the M56 Smartgun. Ready to rumble!")
+		else	to_chat(user, "The powerpack servos finish loading [rounds_to_reload] shells into the M56 Smartgun. Ready to rumble!")
+		playsound(user, 'sound/weapons/unload.ogg', 25, 1)
+
+		reloading = 0
```
Rohesie: FALSE

code/modules/cm_marines/equipment/weapons.dm
```diff
@@ -103,7 +103,43 @@
...
+		return 1
```
Rohesie: TRUE

code/modules/projectiles/updated_projectiles/guns/specialist.dm
```diff
@@ -203,22 +203,25 @@
...
-	var/shells_fired_max = 20 //Smartgun only; once you fire # of shells, it will attempt to reload automatically. If you start the reload, the counter resets.
+	var/shells_fired_max = 50 //Smartgun only; once you fire # of shells, it will attempt to reload automatically. If you start the reload, the counter resets.
 	var/shells_fired_now = 0 //The actual counter used. shells_fired_max is what it is compared to.
 	var/restriction_toggled = 1 //Begin with the safety on.
 	gun_skill_category = GUN_SKILL_SMARTGUN
 	attachable_allowed = list(
+						/obj/item/attachable/extended_barrel,
 						/obj/item/attachable/heavy_barrel,
+						/obj/item/attachable/flashlight,
+						/obj/item/attachable/lasersight,
```
Rohesie: These two offer no trade-off nor are thematically fitting. Don't add stuff just to add it.

Surrealaser: The tradeoff with the laser sight is that you don't get to run a BFA which can actually worth a shit now that the EB is a thing in exchange for a minor accuracy bonus.

Also a rail light is totally thematic, and the benefit is minor given your thermals.

Rohesie:
Let's test this new value.

PR #337, opened 2018-09-19
Refactors(?) coins and tokens interaction with vendors.

Ghommie:
Added in token flags for coins and vendors. They are used to prevent special coins from being inserted in unsupported vendors and special vendors accepting any coin.
This makes coins and future ideaguys projects viable without repercussion of exploits, and may stops a dumb engineer from wasting its token on an empty ol M4A1 or a cigar case. (also fixes a bug or two... or three)

### DIFF
--- a/code/__DEFINES/objects.dm
+++ b/code/__DEFINES/objects.dm
@@ -134,4 +134,4 @@ var/list/RESTRICTED_CAMERA_NETWORKS = list( //Those networks can only be accesse
 #define TOKEN_MARINE			2
 #define TOKEN_ENGI				4
 #define TOKEN_SPEC				8
-#define TOKEN_ALL				16
+#define TOKEN_ALL				15
--- a/code/__DEFINES/objects.dm
+++ b/code/__DEFINES/objects.dm
@@ -127,3 +127,11 @@ var/list/RESTRICTED_CAMERA_NETWORKS = list( //Those networks can only be accesse
 #define IS_PRY_CAPABLE_CROWBAR		2 //actual crowbar
 #define IS_PRY_CAPABLE_FORCE		3 //can force open even powered airlocks
 
+//flags_token & tokensupport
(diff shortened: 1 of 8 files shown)


code/__DEFINES/objects.dm
```diff
@@ -127,3 +127,11 @@ var/list/RESTRICTED_CAMERA_NETWORKS = list( //Those networks can only be accesse
...
 #define IS_PRY_CAPABLE_FORCE		3 //can force open even powered airlocks
 
+//flags_token & tokensupport
+//used for coins and vendors, restricting specific tokens to associated vendors.
+
+#define TOKEN_GENERAL			1
+#define TOKEN_MARINE			2
+#define TOKEN_ENGI				4
+#define TOKEN_SPEC				8
+#define TOKEN_ALL				16
```
Rohesie: This value would be 15 if you want to activate all the former flags.
This is because these are the values in binary:
1: 0001
2: 0010
4: 0100
8: 1000
15: 1111
16: 10000
Look for "bitwise operations" in order to understand how this works.

code/game/machinery/vending/vending.dm
```diff
@@ -168,9 +170,20 @@
...
-			coin = W
-			to_chat(user, "\blue You insert the [W] into the [src]")
+		var/obj/item/coin/C = W
+		if(coin)
+			to_chat(user, "<span class='warning'>[src] already has [coin] inserted</span>")
+			return
+		if(!premium.len && !istype(src,/obj/machinery/vending/shared_vending))
+			to_chat(user, "<span class='warning'>[src] doesn't have a coin slot.</span>")
+			return
+		if(C.flags_token & tokensupport || C.flags_token & TOKEN_ALL)
```
Rohesie: There is no need for this second check if the TOKEN_ALL actually has all bits on as recommended before.

PR #338, opened 2018-09-19
Makes overwatch squad transfers update headset frequency

Jalordon:
Transferred marine's  headsets will now match the new squad's frequency. Only applies to marine type headsets.

### DIFF
--- a/code/modules/cm_marines/overwatch.dm
+++ b/code/modules/cm_marines/overwatch.dm
@@ -767,6 +767,18 @@
 
 	var/obj/item/card/id/ID = transfer_marine.wear_id
 	ID.assigned_fireteam = 0 //reset fireteam assignment
+	
+	//Changes headset frequency to match new squad
+	var/obj/item/device/radio/headset/almayer/marine/H = transfer_marine.wear_ear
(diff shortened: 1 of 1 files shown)


code/modules/cm_marines/overwatch.dm
```diff
@@ -767,6 +767,18 @@
...
+	var/obj/item/device/radio/headset/almayer/marine/H = transfer_marine.wear_ear
+	if(H && istype(H, /obj/item/device/radio/headset/almayer/marine))
+		if (istype(new_squad,/datum/squad/alpha))
+			H.set_frequency(ALPHA_FREQ)	
+		else if (istype(new_squad,/datum/squad/bravo))
+			H.set_frequency(BRAVO_FREQ)		
+		else if (istype(new_squad,/datum/squad/charlie))
+			H.set_frequency(CHARLIE_FREQ)
+		else if (istype(new_squad,/datum/squad/delta))
+			H.set_frequency(DELTA_FREQ)
```
Rohesie: You could simplify this with something like:
	`H.set_frequency(new_squad.radio_freq)`
That would make it squad-agnostic, shorter and using less checks.

PR #340, opened 2018-09-20
Fixes heart damage simply making your chest go doki doki.

Ghommie:
downstream fix of a bug reported on the GitLab, and slight tweaks .
And reminder to avoid unregularized assumption of hyperzine and keep the dosages low, gamers.

### DIFF
--- a/code/modules/mob/living/blood.dm
+++ b/code/modules/mob/living/blood.dm
@@ -49,7 +49,7 @@
 				if(heart.damage < heart.min_bruised_damage)
 					heart_multi = 0.9
 					blood_volume = max(blood_volume - 0.1, 0) //nulls regeneration
-				else if(heart.damage >= heart.min_bruised_damage && heart.damage < heart.min_broken_damage)
+				else if(heart.damage < heart.min_broken_damage)
 					heart_multi = 0.7
 					blood_volume = max(blood_volume - 0.5, 0)
 				else
--- a/code/modules/mob/living/blood.dm
+++ b/code/modules/mob/living/blood.dm
@@ -45,14 +45,14 @@
 			if(!heart)
 				heart_multi *= 0.5 //you'd die in seconds but you can't remove internal organs even with varediting.
 
-			if((!reagents.get_reagent_amount("peridaxon") >= 0.05) && heart.damage > 1)
+			if(!(reagents.get_reagent_amount("peridaxon") >= 0.05) && heart.damage > 1)
 				if(heart.damage < heart.min_bruised_damage)
 					heart_multi = 0.9
 					blood_volume = max(blood_volume - 0.1, 0) //nulls regeneration
 				else if(heart.damage >= heart.min_bruised_damage && heart.damage < heart.min_broken_damage)
 					heart_multi = 0.7
 					blood_volume = max(blood_volume - 0.5, 0)
-				else if(heart.damage >= heart.min_broken_damage && heart.damage < INFINITY)
+				else
 					heart_multi = 0.5
 					blood_volume = max(blood_volume - 1.3, 0)
 			else
--- a/code/modules/mob/living/blood.dm
+++ b/code/modules/mob/living/blood.dm
@@ -5,6 +5,7 @@
 
 /mob/living
 	var/blood_volume = 0 //how much blood the mob has
+	var/heart_multi = 1
 
 /mob/living/carbon
 	blood_volume = BLOOD_VOLUME_NORMAL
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/blood.dm
```diff
@@ -36,50 +37,56 @@
...
+			else
+				heart_multi = 1
 
 
 
 
 
 	//Effects of bloodloss
-		switch(blood_volume)
+		switch(blood_volume / heart_multi)
```
Rohesie: Can you change the `heart_multi` logic to avoid a division here? Multiplying `blood_volume` by 0.5 would achieve the same result than dividing by 2, for example.

Ghommie:
Done so.

code/modules/mob/living/blood.dm
```diff
@@ -36,50 +37,56 @@
...
-				b_volume = 0
-			else if(reagents.get_reagent_amount("peridaxon") >= 0.05)
-				b_volume *= 1
-			else if(heart.damage > 1 && heart.damage < heart.min_bruised_damage)
-				b_volume *= 0.8
-			else if(heart.damage >= heart.min_bruised_damage && heart.damage < heart.min_broken_damage)
-				b_volume *= 0.6
-			else if(heart.damage >= heart.min_broken_damage && heart.damage < INFINITY)
-				b_volume *= 0.3
+				heart_multi += 1 //you'd die in seconds but you can't remove internal organs even with varediting.
```
Rohesie: I'd assume that under the new logic you'd want this to be something like `heart_multi = 0.5`, as increasing the value here would actually increase the blood volume further down. Or something harsher, given that you've changed the values of the other ones.

code/modules/mob/living/blood.dm
```diff
@@ -36,50 +37,56 @@
...
+			if((!reagents.get_reagent_amount("peridaxon") >= 0.05) && heart.damage > 1)
+				if(heart.damage < heart.min_bruised_damage)
+					heart_multi = 0.9
+					blood_volume -= 0.1 //nulls regeneration
+				else if(heart.damage >= heart.min_bruised_damage && heart.damage < heart.min_broken_damage)
+					heart_multi = 0.7
+					blood_volume -= 0.5
+				else if(heart.damage >= heart.min_broken_damage && heart.damage < INFINITY)
+					heart_multi = 0.5
+					blood_volume -= 1.3
```
Rohesie: Couldn't this end up in a negative blood amount?

Ghommie: Handle blood doesn't process when you are dead on in cryo.

Rohesie: I don't think this is a big issue, but it's still good practice. Everywhere else that is uncommented in the code does something similar to `blood_volume = max(0, blood_volume - amount)`. Might want to implement something like that.

code/modules/mob/living/blood.dm
```diff
@@ -36,50 +37,56 @@
...
-				b_volume *= 0.3
+				heart_multi *= 0.5 //you'd die in seconds but you can't remove internal organs even with varediting.
+
+			if((!reagents.get_reagent_amount("peridaxon") >= 0.05) && heart.damage > 1)
+				if(heart.damage < heart.min_bruised_damage)
+					heart_multi = 0.9
+					blood_volume -= max(0, blood_volume - 0.1) //nulls regeneration
+				else if(heart.damage >= heart.min_bruised_damage && heart.damage < heart.min_broken_damage)
+					heart_multi = 0.7
+					blood_volume = max(0, blood_volume - 0.5)
```
Rohesie: This should be `blood_volume -= max(0, blood_volume - 0.5)` unless if we want to drain people's blood rather fast.

code/modules/mob/living/blood.dm
```diff
@@ -36,50 +37,56 @@
...
+				else if(heart.damage >= heart.min_broken_damage && heart.damage < INFINITY)
+					heart_multi = 0.5
+					blood_volume = max(0, blood_volume - 1.3)
```
Rohesie: Yet another case of evil vampirism.

code/modules/mob/living/blood.dm
Ghommie: hold on, are you sure about that? To my logic that means that if our volume is, let's say, 400 or something and the organ bruised, we'd end up having blood -= max(0, 399.5)... may be wrong, tho.

code/modules/mob/living/blood.dm
```diff
@@ -36,50 +37,56 @@
...
-				b_volume *= 0.6
-			else if(heart.damage >= heart.min_broken_damage && heart.damage < INFINITY)
-				b_volume *= 0.3
+				heart_multi *= 0.5 //you'd die in seconds but you can't remove internal organs even with varediting.
+
+			if(!(reagents.get_reagent_amount("peridaxon") >= 0.05) && heart.damage > 1)
+				if(heart.damage < heart.min_bruised_damage)
+					heart_multi = 0.9
+					blood_volume = max(blood_volume - 0.1, 0) //nulls regeneration
+				else if(heart.damage >= heart.min_bruised_damage && heart.damage < heart.min_broken_damage)
```
Rohesie: The first check is not necessary. The `else` can only be triggered if `heart.damage` is not lower than `heart.min_bruised_damage`, and thus equal or higher.

PR #343, opened 2018-09-20
More consistency and exploit patches for the autolathe.

Ghommie:
After a general budget inspection, the uscm opted to replace faux wood metal stocks (that could have been decon'd in the autolathe for metal) with authentic ones.
The combat flashlight is a little more "combat" now.
Also it's no more possible to generate infinite metal with rail lights/flashlight conversions, nor you can convert lanterns.
Slight tweak with combat knife/bayonet matter amounts.

### DIFF
--- a/code/game/objects/items/devices/flashlight.dm
+++ b/code/game/objects/items/devices/flashlight.dm
@@ -7,9 +7,7 @@
 	w_class = 2
 	flags_atom = FPRINT|CONDUCT
 	flags_equip_slot = SLOT_WAIST
-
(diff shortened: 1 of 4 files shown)


code/game/objects/items/devices/flashlight.dm
```diff
@@ -297,3 +295,4 @@
 	icon_state = "lantern"
 	desc = "A mining lantern."
 	brightness_on = 6			// luminosity when on
+	raillight_compatible = 0
```
Rohesie: FALSE

PR #345, opened 2018-09-20
no more Methuselah marines and crewcut oppression.

Ghommie:
The maximum age for marines has been reduced from the ultra-eldery 160 years to 85 (/tg/ standard).
New Clients' chars don't necessarily start with the standard crewcut anymore, instead it picks one from a tiny, boring list. Decent odds they'll be baldie.
Lil randomization on the age. Like if that even matters anyway.

Open to opinions and criticism.


Rohesie:
I personally prefer bald male to be the sole default so newbies (and, unfortunately, lazy griefers) can be more easily identified.
That said, the change is otherwise fine with me.

PR #348, opened 2018-09-21
Enables squad marine access to general chat

Kingtrin:
babies first pr

This gives squad marines access to the almayer general chat. Some discussion was had on whether general chat should be default, but I feel that doing so would mean people would not communicate with their squad as often. The squad encryption keys (these weren't actually used anywhere) were used to achieve this effect, and so each marine headset now comes with at least one key. This should hopefully be compatible with #338 because if it isn't imma cri.

Descriptions were changed on each headset affected. The keybind for general chat, when it isn't the default frequency, is :Z.

Apologies for the ugly commits I'm still figuring out how to git.

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -71,7 +71,7 @@ datum/mind
 
 
 
-	proc/transfer_to(mob/living/new_character)
+	proc/transfer_to(mob/living/new_character, var/force_key_move = FALSE)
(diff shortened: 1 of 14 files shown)


Rohesie:
This gives squad marines access to the Almayer general chat.
Descriptions were changed on each headset affected. The keybind for general chat, when it isn't the default frequency, is :Z.

PR #355, opened 2018-09-22
Barricade Bypass Update

Surrealaser:
-IFF bullets now have no barricade block chance.

-M56D bullets can travel up to two tiles before having a barricade block chance, slightly improving their range of viable placement.

### DIFF
--- a/code/modules/projectiles/updated_projectiles/projectile.dm
+++ b/code/modules/projectiles/updated_projectiles/projectile.dm
@@ -330,9 +330,7 @@
 	if(!( P.dir & reverse_direction(dir) || P.dir & dir))
 		return FALSE //no effect if bullet direction is perpendicular to barricade
 
-	var/distance = P.distance_travelled - 1
-
-	if(P.ammo.barricade_clear_distance > distance)
+	if(P.distance_travelled <= P.ammo.barricade_clear_distance)
 		return FALSE
 
 	var/coverage = 90 //maximum probability of blocking projectile
--- a/code/modules/projectiles/updated_projectiles/projectile.dm
+++ b/code/modules/projectiles/updated_projectiles/projectile.dm
@@ -321,7 +321,7 @@
 	if(!throwpass)
 		return TRUE
 
-	if(P.ammo.flags_ammo_behavior & AMMO_SNIPER) //sniper rounds bypass cover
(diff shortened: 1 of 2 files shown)


code/modules/projectiles/updated_projectiles/projectile.dm
```diff
@@ -332,7 +332,7 @@
 
 	var/distance = P.distance_travelled - 1
 
-	if(distance < 1)
+	if(P.ammo.barricade_clear_distance > distance)
```
Rohesie: To follow your logic here you want a `=>` instead.

Surrealaser: ???

Default clear distance allowance is 1.

Distance traveled is always distance traveled - 1.

Most projectiles are intended to ignore barricades 1 tile away, so this works: their barricade clear distance will be 1 and the distance after mods, will be 0. 1 is greater than 0 so the bypass will trigger.


Or did you want this to be reformatted, such that distance = P.distance traveled with >= used instead?

Rohesie: Ah, I missed the -1 there. Not sure why it's there in any case, you could simplify it with `if(P.distance_travelled <= P.ammo.barricade_clear_distance)`.

Surrealaser: I was following existing formatting.

Rohesie: All good. In this case `distance` is a variable and `P.ammo.barricade_clear_distance` a number. Please alter it to the `variable` `operator` `number` format as per the contribution guide (that would mean a `if(distance < P.ammo.barricade_clear_distance)`)

code/modules/projectiles/updated_projectiles/projectile.dm
```diff
@@ -321,18 +321,17 @@
...
 		return FALSE //window frames, unflipped tables
 
 	if(!( P.dir & reverse_direction(dir) || P.dir & dir))
 		return FALSE //no effect if bullet direction is perpendicular to barricade
-
-	var/distance = P.distance_travelled - 1
-
-	if(distance < 1)
+		
+	var/distance = P.distance_travelled
```
Rohesie: This change will have an effect in `var/hitchance`. Are you taking it into account?
You could also delete this unnecessary var if you substituted `distance` with `P.distance_travelled` down there. Just note that before your changes that was `P.distance_travelled - 1`

PR #357, opened 2018-09-22
Improved Revolvers/Pistols

Surrealaser:
Significantly increased the fire speed of revolvers and pistols so they're worth at least half a shit.


Supermichael777:
can you make the revolvers unique action open the cylinder?

PR #360, opened 2018-09-22
Tackle Rework for Xenos

Surrealaser:
-Tackling is now guaranteed. However, it only knocks down enemies for a very short duration, and only after they've accumulated sufficient pain (100+), and haven't been stunned or knocked down in the past 2 ticks.

-Tackling will not knockdown someone who is already knocked down.

-Tackling will always apply armor ignoring Halloss damage (amount depends on caste/upgrade level; Frenzy increases this). Each unit of Halloss damage causes 1.5 units of pain that cannot be reduced by pain killers. If total pain exceeds 200, then a human will be instantly disabled for a significant duration; this is the primary mechanism by which marines can be captured.

-Also added a cooldown on RNG crits while I was in attack_alien: one per 2.5 seconds max.


Supermichael777:
Uhhhh
https://github.com/ColonialMarines-Mirror/ColonialMarines-2018/blob/08d1ede8d07aa7dd53458d8dead3236e059754c6/code/modules/mob/living/carbon/human/life/handle_regular_status_updates.dm#L34

Do a basic search on how its handled and reduced?

ElfLeaderMike:
I get why this was done, but after playing with it it pretty much ruined the ability to play runner.
I landed 8 successful tackles between two people and couldn't even juggle the both of them for kills or captures, it makes runner no longer fun to play.

-1 because honestly it ruins my runner main life and in practice it feels just as shitty as CM

dasWurmtMich:
I know why its being done but it really needs adjusting, the main stun should be atleast somewhat longer.

The problem about the previous tackle system was that stuns just lasted a ridiculous amount of time for some castes (like drones...) but rightnow T1 tackle ability got demolished. Even elite runners and upwards can't really tackle. Personally I think the previous tackle system was okay but the stun duration was far too long, now the stun duration is far too short. There has to be a mid point in terms of stun duration.

Also when queen screech is getting reworked/gutted xenos will have a hard time tackling multiple marines

Surrealaser:
I heard a warrior was only stutterstunning after 8 tackles or so with 40 halloss damage; if this is the case it is not intended behaviour; I'll look over the code and see what needs adjusting.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -211,7 +211,12 @@
 			AdjustKnockedout(-3)
 			AdjustStunned(-3)
 			AdjustKnockeddown(-3)
-
+			if(halloss > 0)
+				if(ishuman(M))
```
Rohesie: Since this affects only humans you'll want it under `/mob/living/carbon/human/proc/handle_regular_status_updates()`

code/modules/mob/living/carbon/carbon_defines.dm
```diff
@@ -3,6 +3,8 @@
 	var/datum/species/species //Contains icon generation and language information, set during New().
 	var/list/stomach_contents = list()
 
+	var/base_halloss_recovery_rate = -2 //Points of Halloss recovered per life cycle while awake.
+	var/rest_halloss_recovery_rate = -10 //Points of Halloss recovered per life cycle while sleeping/resting/unconscious.
```
Rohesie: This would go under human_defines.dm

code/modules/mob/living/carbon/human/life/handle_regular_status_updates.dm
```diff
@@ -158,11 +151,16 @@
...
 		marskman_aura = marskman_new
 
-		//hud_set_pheromone() //TODO: HOOK THIS UP, ASK PHIL
-
 		mobility_new = 0
 		protection_new = 0
 		marskman_new = 0
+		
+		//Natural recovery; enhanced by hold/protection aura.
+		var/aura_recovery_multiplier = 1 + max(0,0.5 * protection_aura) //Protection aura adds +50% recovery rate per point of leadership beyond the first; +150% for a CO/XO
```
Rohesie: Make this recovery mirror the frenzy bonus to cancel it out.

code/modules/mob/living/carbon/human/life/handle_shock.dm
```diff
@@ -4,14 +4,25 @@
...
 
-	if(health < config.health_threshold_softcrit) 		shock_stage = max(shock_stage, 60)//If they took too much damage, they immediately enter shock.
+	if(health < config.health_threshold_softcrit)
+		shock_stage = max(shock_stage, 60)//If they took too much damage, they immediately enter shock.
 
-	if(traumatic_shock >= 80) 							shock_stage++ //If they shock exceeds 80, add more to their shock stage, regardless of health.
-	else if(health < config.health_threshold_softcrit) 	shock_stage = max(shock_stage, 60)
-	/*If their health is lower than threshold, but they don't have enough shock, they will never go below 60.
-	Otherwise they slowly lose shock stage.*/
+	if(traumatic_shock > 200)
```
Rohesie: Use a `switch`

code/modules/mob/living/carbon/xenomorph/attack_alien.dm
```diff
@@ -80,13 +80,15 @@
 
 			//Check for a special bite attack
 			if(prob(M.bite_chance))
-				M.bite_attack(src, damage)
-				return 1
+				if(!M.critical_proc) //Can't crit if we already crit in the past 3 seconds
+					M.bite_attack(src, damage)
+					return 1
```
Rohesie: `TRUE`

code/modules/mob/living/carbon/xenomorph/attack_alien.dm
```diff
@@ -80,13 +80,15 @@
...
 
 			//Check for a special bite attack
 			if(prob(M.tail_chance))
-				M.tail_attack(src, damage)
-				return 1
+				if(!M.critical_proc) //Can't crit if we already crit in the past 3 seconds
+					M.tail_attack(src, damage)
+					return 1
```
Rohesie: `TRUE`

code/modules/mob/living/carbon/human/life/handle_regular_status_updates.dm
Surrealaser: Frenzy bonus is a bonus to the damage.

Protection aura provides a bonus to the recovery rate.

These things can't really be the same, unless you want the protection aura to reduce the incoming tackle damage by the same amount as the Frenzy.

code/modules/mob/living/carbon/carbon.dm
Surrealaser: I'm forward thinking on this in case we have antags/others that halloss interactions might apply to in the future.

Rohesie: They are more likely to be a subset of humans, such as synths are...

code/modules/mob/living/carbon/carbon_defines.dm
Rohesie: They are also constants, are they not?

Rohesie:
This is a controversial but necessary change. Those who are downvoting, please offer alternatives or changes to be made besides leaving everything the same.

The idea is to move away from stun based combat, more into disablers than tasers, for reference.

code/modules/mob/living/carbon/human/human_defines.dm
```diff
@@ -118,3 +118,6 @@
 	var/temporary_slowdown = 0 //Stacking slowdown caused from effects, currently used by neurotoxin gas
 
 	var/datum/squad/assigned_squad //the squad this human is assigned to
+
+	var/base_halloss_recovery_rate = -2 //Points of Halloss recovered per life cycle while awake.
+	var/rest_halloss_recovery_rate = -10 //Points of Halloss recovered per life cycle while sleeping/resting/unconscious.
```
Rohesie: The rest of the variables in this file are things that change, variables.
Since you are using constant values you can define them as such, constants.

code/modules/mob/living/carbon/xenomorph/Castes/Crusher.dm
```diff
@@ -6,9 +6,8 @@
 	icon_state = "Crusher Walking"
 	melee_damage_lower = 15
 	melee_damage_upper = 30
-	tacklemin = 4
-	tacklemax = 7
-	tackle_chance = 60
+	tackle_damage = 55
+	attack_delay = 0.5
```
Rohesie: The crusher is big and slow. Give it a strong tackle, but not a quick one, please.

code/modules/mob/living/carbon/xenomorph/attack_alien.dm
```diff
@@ -155,49 +157,35 @@
...
-						"<span class='danger'>You try to tackle [src]</span>", null, 5)
-						return 1
-				else if(prob(M.tackle_chance + tackle_bonus)) //Tackle_chance is now a special var for each caste.
-					playsound(loc, 'sound/weapons/alien_knockdown.ogg', 25, 1)
-					KnockDown(rand(M.tacklemin, M.tacklemax))
-					M.visible_message("<span class='danger'>\The [M] tackles down [src]!</span>", \
-					"<span class='danger'>You tackle down [src]!</span>", null, 5)
-					return 1
 
+			if(knocked_down)
```
Rohesie: Being unable to tackle knocked down targets makes for very hard keeping of hosts. You'll might want something like `if(knocked_out || sleeping)`

code/modules/mob/living/carbon/xenomorph/Castes/Crusher.dm
Surrealaser: This actually makes it attack slower.

Negative values = faster.

code/modules/mob/living/carbon/xenomorph/attack_alien.dm
Surrealaser: The entire point is to prevent chain stuns.

I've already given the duration a bonus that scales with halloss.

Rohesie: It defeats the purpose of the stamina-like damage if it cannot stack. It already has a max value it can reach.

Surrealaser: Halloss is supposed to stack.

Hard CC is not suppose to stack; stuns are not supposed to chain. This is half the fundamental point of this PR.

code/modules/mob/living/carbon/xenomorph/attack_alien.dm
```diff
@@ -155,49 +157,35 @@
...
 				playsound(loc, 'sound/weapons/alien_claw_swipe.ogg', 25, 1)
-				M.visible_message("<span class='danger'>\The [M] tries to tackle [src]</span>", \
-				"<span class='danger'>You try to tackle [src]</span>", null, 5)
+				M.visible_message("<span class='danger'>\The [M] tries to tackle [src], but they are already down!</span>", \
+				"<span class='danger'>You try to tackle [src], but they are already down!</span>", null, 5)
+			else
+				playsound(loc, 'sound/weapons/alien_knockdown.ogg', 25, 1)
+				var/knockdown_duration = traumatic_shock * 0.005 
+				KnockDown(max(0.3 + knockdown_duration)) //Highest of 0.3 seconds or 0.5% of total pain.
```
Rohesie: This looks like it could use a comma instead of a plus sign.

code/modules/mob/living/carbon/xenomorph/attack_alien.dm
```diff
@@ -155,49 +157,35 @@
...
+			apply_damage(tackle_pain, HALLOSS)
+			updatehealth()
+			var/throttle_message = "The [M] throttles [src]!"
+			var/throttle_message2 = "You throttle [src]!</span>"
+			if(tackle_pain > 40)
+				throttle_message = "<span class='danger'>\The [M] throttles [src]!</span>"
+				throttle_message2 = "<span class='danger'>You badly throttle [src]!</span>"
+			M.visible_message("[throttle_message]", \
+			"[throttle_message2]", null, 5)
+			return 1
```
Rohesie: `TRUE`

code/modules/mob/living/carbon/xenomorph/attack_alien.dm
Surrealaser: FUCK

Rohesie:
What's with the xeno health changes? That falls outside the scope of this PR.

Surrealaser:
@Rohesie That's a product of the rebase

code/modules/mob/living/carbon/human/life/handle_shock.dm
```diff
@@ -30,34 +44,39 @@
...
-			if(shock_stage == 60 && !lying)
-				emote("me", 1, " is having trouble standing.")
 			eye_blurry = max(2, eye_blurry)
 			stuttering = max(stuttering, 5)
 			if(prob(2))
+				if(!lying)
+					emote("me", 1, " is having trouble standing.")
 				to_chat(src, "<span class='danger'>[pick("The pain is excruciating", "Please, just end the pain", "Your whole body is going numb")]!</span>")
-				KnockDown(20)
+				KnockDown(5)
```
Rohesie: This is a big change to how pain is processed, not just halloss. Will require some testing.

code/modules/mob/living/carbon/xenomorph/attack_alien.dm
Rohesie: I understand not stacking `knocked_down`, that is good, but `halloss` should stack, and the xeno should keep adding it to a conscious target if they keep attacking. That's how disabler damage works.

Rohesie: You can add a check to see if the target is unconscious and then add the "but they are already down!" message.
If the target is not receiving `knockdown` but is still receiving `halloss` then a message denoting it should be added.

code/modules/mob/living/carbon/human/life/handle_shock.dm
```diff
@@ -7,25 +7,25 @@
...
 			return
 
 		if(74 to 50)
-			shock_stage = max(0, min(shock_stage - 2, 80))
+			shock_stage = CLAMP(shock_stage - 2, 0, 80))
 			return
 
 		if(49 to 0)
-			shock_stage = max(0, min(shock_stage - 10, 60)) //When we have almost no pain remaining. No greater than 60 and no smaller than 0, reduced by 10 each time.
+			shock_stage = CLAMP(shock_stage - 10, 0, 60)) //When we have almost no pain remaining. No greater than 60 and no smaller than 0, reduced by 10 each time.
```
Rohesie: Why reducing the `shock_stage` so fast when we may still have `traumatic_shock`?
Go a little softer on the recovery.
This will definitely require some playtesting.

Surrealaser: Wait what?

49-0 is literally the traumatic shock they have, which is quite little.

code/modules/mob/living/carbon/human/life/handle_shock.dm
```diff
@@ -2,28 +2,27 @@
 
 /mob/living/carbon/human/handle_shock()
 	..()
-	if(status_flags & GODMODE || analgesic || (species && species.flags & NO_PAIN)) return //Godmode or some other pain reducers. //Analgesic avoids all traumatic shock temporarily
+	if(status_flags & GODMODE || analgesic || (species && species.flags & NO_PAIN) || stat == DEAD || !traumatic_shock)
```
Rohesie: `!traumatic_shock` means `shock_stage` could be higher than 0 and never diminish until the creature is hurt again.

code/modules/mob/living/carbon/human/life/handle_shock.dm
```diff
@@ -2,28 +2,27 @@
...
+		return //Godmode or some other pain reducers. //Analgesic avoids all traumatic shock temporarily
 
-	switch(CLAMP(traumatic_shock, 0, 10000))
-		if(10000 to 200)
-			shock_stage = CLAMP(shock_stage++,150,200) //Indescribable pain. At this point, you will immediately be knocked down, with shock stage set to 150.
-			return
+	switch(CLAMP(traumatic_shock,0,400))
+		if(201 to 400)
```
Rohesie: You would get the same behavior by clamping it to 201 and then running an `if(201)`

code/modules/mob/living/carbon/xenomorph/attack_alien.dm
```diff
@@ -168,23 +168,22 @@
...
-			var/tackle_pain = null
-			tackle_pain = (M.tackle_damage * rand(0.5,1.5) + M.tackle_damage * rand(0.5,1.5)) * 0.5 //Gaussian distribution
+			playsound(loc, 'sound/weapons/alien_knockdown.ogg', 25, 1)
+			var/tackle_pain = (rand(M.tackle_damage * 0.75 , M.tackle_damage * 1.25) + rand(M.tackle_damage * 0.75 , M.tackle_damage * 1.25)) * 0.5
 			if(M.frenzy_aura)
-				tackle_pain *= 1 + (0.05 * M.frenzy_aura)  //Halloss damage increased by 5% per rank of frenzy aura
+				tackle_pain = tackle_pain * (1 + (0.05 * M.frenzy_aura))  //Halloss damage increased by 5% per rank of frenzy aura
 			if(protection_aura)
-				tackle_pain *= 1 - (0.10 + 0.05 * protection_aura)  //Halloss damage decreased by 10% + 5% per rank of protection aura
+				tackle_pain = tackle_pain * (1 - (0.10 + 0.05 * protection_aura))  //Halloss damage decreased by 10% + 5% per rank of protection aura
```
Rohesie: Why this reduction?

code/__DEFINES/mobs.dm
```diff
@@ -438,6 +438,7 @@ var/list/global_mutations = list() // list of hidden mutation things
 #define BASE_HALLOSS_RECOVERY_RATE -2
 #define REST_HALLOSS_RECOVERY_RATE -10
 
+
```
Rohesie: Try to clean your code, not leaving unintended modifications.

code/modules/mob/living/carbon/human/life/handle_shock.dm
```diff
@@ -30,34 +42,39 @@
...
+				if(!lying)
+					emote("me", 1, " is having trouble standing.")
 				to_chat(src, "<span class='danger'>[pick("The pain is excruciating", "Please, just end the pain", "Your whole body is going numb")]!</span>")
-				KnockDown(20)
+				KnockDown(5)
 			if(prob(2))
 				to_chat(src, "<span class='danger'>[pick("You black out", "You feel like you could die any moment now", "You're about to lose consciousness")]!</span>")
 				KnockOut(5)
-		if(150 to INFINITY)
+		if(10000 to 150)
```
Rohesie: INFINITY should be fine to use.

code/modules/mob/living/carbon/human/life/handle_shock.dm
Surrealaser: Urgh, was left over from a test.

code/modules/mob/living/carbon/human/life/handle_shock.dm
Surrealaser: I'm aware, I clamped it to the likely highest attainable level during testing; this works fine though.

code/modules/mob/living/carbon/xenomorph/attack_alien.dm
Surrealaser: I'm pretty sure you requested the Hold Aura work to counteract the impact of Frenzy.

I thought it was a good idea and made sense so I included it.

code/__DEFINES/mobs.dm
Surrealaser: ?

Rohesie: Check the changes. Your PR is adding a whitespace in this file. It's clearly something you added and then removed, leaving a space. Try to clean after your changes.

Rohesie:
Solve the conflict issues so we can test this again.

Rohesie:
The merge-garbage needs to be cleaned before we can test this again.

code/modules/mob/living/carbon/human/human_attackhand.dm
```diff
@@ -235,4 +235,45 @@
...
\ No newline at end of file
+
+		else
+			var/t_him = "it"
+			if (gender == MALE)
+				t_him = "him"
+			else if (gender == FEMALE)
+				t_him = "her"
+			if (w_uniform)
+				w_uniform.add_fingerprint(M)
```
Rohesie: Garbage?

code/modules/mob/living/carbon/human/human_attackhand.dm
```diff
@@ -235,4 +235,5 @@
 				to_chat(src, "\t [status=="OK"?"\blue ":"\red "]My [org.display_name] is [status].")
 			if((SKELETON in mutations) && !w_uniform && !wear_suit)
 				play_xylophone()
+
```
Rohesie:
```suggestion
```

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -228,15 +241,12 @@
...
-			AdjustKnockedout(-3)
-			AdjustStunned(-3)
-			AdjustKnockeddown(-3)
-
 			playsound(src.loc, 'sound/weapons/thudswoosh.ogg', 25, 1, 5)
 
 
 
 
+
```
Rohesie:
```suggestion
```

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -220,6 +220,19 @@
...
+				AdjustKnockedout(-3)
+				AdjustStunned(-3)
+				AdjustKnockeddown(-3)
+
+				if(halloss > 0)
+					var/halloss_mod = 1
+					if(ishuman(src))
+						var/mob/living/carbon/human/H = src
+						if(H.protection_aura)
+							halloss_mod = 0.5 + 0.5 * H.protection_aura //SL Hold Aura bonus: +100% of the normal recovery bonus, SO: +150%, XO/CO: +200%
```
Jalordon: +=  to match the intent of the comment

code/modules/mob/living/carbon/human/life/handle_shock.dm
```diff
@@ -2,66 +2,99 @@
 
 /mob/living/carbon/human/handle_shock()
 	..()
```
Rohesie:
```suggestion
	. = ..()
```

PR #366, opened 2018-09-23
Xenos Can't Drag Corpses

Surrealaser:
-Prevents xenos from dragging corpses.

-Xenos auto-drop mobs that die.

-Xenos can 'attack' corpses with lights to disable those lights.

### DIFF
--- a/code/modules/mob/death.dm
+++ b/code/modules/mob/death.dm
@@ -43,10 +43,6 @@
 	if(!gibbed)
 		src.visible_message("<b>\The [src.name]</b> [deathmessage]")
 
-	if(pulledby)
-		if(isXeno(pulledby)) //Xenos drop dead guys
-			pulledby.stop_pulling()
-
 	stat = DEAD
 
 	update_canmove()
--- a/code/modules/mob/death.dm
+++ b/code/modules/mob/death.dm
@@ -43,6 +43,10 @@
 	if(!gibbed)
 		src.visible_message("<b>\The [src.name]</b> [deathmessage]")
 
+	if(pulledby)
+		if(isXeno(pulledby)) //Xenos drop dead guys
+			pulledby.stop_pulling()
+
 	stat = DEAD
 
 	update_canmove()
--- a/code/modules/mob/living/carbon/xenomorph/attack_alien.dm
+++ b/code/modules/mob/living/carbon/xenomorph/attack_alien.dm
@@ -30,7 +30,12 @@
 				return 0
 
 			if(Adjacent(M)) //Logic!
-				M.start_pulling(src)
+				if(stat == DEAD)
+					if(is_revivable())
+						to_chat(src, "<span class='warning'>The creature is dead, why would you want to touch it?</span>")
+						return
+				else
+					M.start_pulling(src)
 
 		if("hurt")
 			var/datum/hive_status/hive
--- a/code/modules/mob/living/carbon/xenomorph/attack_alien.dm
+++ b/code/modules/mob/living/carbon/xenomorph/attack_alien.dm
@@ -43,7 +43,18 @@
 				return FALSE
 
 			if(stat == DEAD)
-				to_chat(M, "<span class='warning'>[src] is dead, why would you want to touch it?</span>")
+				if(luminosity > 0)
+					playsound(loc, "alien_claw_metal", 25, 1)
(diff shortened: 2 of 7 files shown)


Supermichael777:
marine lights auto go out on death/other things right now

anything86:
This is not first contact, so I don't know why people would still complain about xenos dragging away bodies. It can alert other marines that someone died there, be looted for extra equipment and in the worst case scenario be used to revive whoever you killed. The justification in CM was that it was meta, with the new rules of engagement I can't see how it would be.

code/modules/mob/living/carbon/xenomorph/attack_alien.dm
```diff
@@ -43,7 +43,18 @@
...
-				to_chat(M, "<span class='warning'>[src] is dead, why would you want to touch it?</span>")
+				if(luminosity > 0)
+					playsound(loc, "alien_claw_metal", 25, 1)
+					M.flick_attack_overlay(src, "slash")
+					var/datum/effect_system/spark_spread/spark_system2
+					spark_system2 = new /datum/effect_system/spark_spread()
+					spark_system2.set_up(5, 0, src)
+					spark_system2.attach(src)
+					spark_system2.start(src)
+					disable_lights()
```
Rohesie: Remove the other instances of this proc, such as when someone is facehugged or dies.

code/modules/mob/mob.dm
```diff
@@ -369,6 +369,13 @@ var/list/slot_equipment_priority = list( \
 	var/mob/M
 	if(ismob(AM))
 		M = AM
+		if(isXeno(src) && M.stat == DEAD)
+			if(ishuman(M))
+				var/mob/living/carbon/human/H = M
+				if(H.is_revivable())
+					to_chat(src, "<span class='warning'>The creature is dead, why would you want to touch it?</span>")
+					return
```
Rohesie: This can be handled in the attack_alien() proc, earlier.

code/modules/mob/death.dm
```diff
@@ -43,6 +43,10 @@
 	if(!gibbed)
 		src.visible_message("<b>\The [src.name]</b> [deathmessage]")
 
+	if(pulledby)
+		if(isXeno(pulledby)) //Xenos drop dead guys
+			pulledby.stop_pulling()
```
Rohesie: This is already handled in the more specific `/mob/living/carbon/human/death`

Rohesie:
I've never minded it myself, but people complained runners with sonic speed would drag defibable bodies to the edge of the map, or under some sprite, so they wouldn't be brought back. It's one of the deadchat salt generators.

CM solved this by forbidding from moving dead bodies that don't have a living larva in it, besides turning down the lights in the body when the victim gets facehugged, dies or anyone equips anything on them. As a result xenos infect so they can drag away to the edge of the map still, and people who get nested have a hard time because they've even lost their lights. The lights automatically turning off is quite annoying and it's a generally messy solution. This one is much simpler and elegant in my eyes.

I took active part in this PR and it involves xeno balance, so I'll leave it to @ninjanomnom to review it.

ninjanomnom:
I'm very against disallowing dragging in all cases. Slowing them down while dragging would be fine but this is unnecessary imo.

Surrealaser:
@ninjanomnom The problem is that merely slowing the drag doesn't actually address the fundamental problem of corpses being dragged away into a pile in some remote corner in order to run out the defib timer.

This system works fine on CM where it's currently in use.

Rohesie:
The dragging restriction is only there for defibable bodies, by the byes.

ninjanomnom:
I still find it very silly to be blocking that sort of thing. Just making it harder to accomplish means they can't do it in combat easily and after combat is over they should be free to bring the bodies to a more secure location.

Surrealaser:
@ninjanomnom Crusher can do it with casual easy.

Rohesie:
Either add a move delay when pulling defibable humans, in `/mob/living/carbon/Xenomorph/movement_delay()` , or limit this PR to removing the lights turning off automatically and adding the lights removal attack, and I'll do it myself in my PR. Let me know what you decide.

Rohesie:
I've added the humans-drag slowdown, as suggested, to #395 . Remove the drag block from this PR to comply with @ninjanomnom 's request and keep the light-related changes so we can merge this.

PR #374, opened 2018-09-24
The longstreet tank is no longer damaged roundstart. 

Tupinambis:
Just kit it our with whatever parts you want for that round and you are good to go.
No more tedious repairs!

[QOL]: The decrepit tank that spawns in the Almayer's lower deck is replaced with a non damaged tank that just needs its hardpoints filled to function.


Rohesie:
Could you resolve the conflicts so we may test this?

PR #377, opened 2018-09-24
Sanitize Chemistry and Sleeper

Rohesie:
Fixes against HTML injection exploits.


code/game/machinery/Sleeper.dm
```diff
@@ -119,30 +119,37 @@
...
+		user.set_interaction(src)
 		if (href_list["chemical"])
-			if (src.connected)
-				if (src.connected.occupant)
-					if (src.connected.occupant.stat == DEAD)
-						to_chat(usr, "\red \b This person has no life for to preserve anymore. Take them to a department capable of reanimating them.")
-					else if(src.connected.occupant.health > 0 || href_list["chemical"] == "inaprovaline")
-						src.connected.inject_chemical(usr,href_list["chemical"],text2num(href_list["amount"]))
+			if (connected)
+				if (connected.occupant)
```
ninjanomnom: Try to reduce indentation and save on white space and combine your conditionals if it doesn't make it confusing

code/modules/reagents/Chemistry-Machinery.dm
```diff
@@ -241,7 +244,7 @@
 	use_power = 1
```
Qustinnus: not sure if this is a bool, but theres a few bools on this object that dont use defines

Rohesie: No, this is not exactly a boolean. `use_power` doesn't use defines, though you are right that it should.

PR #378, opened 2018-09-24
getarea proc from recursion to iteration.

Rohesie:
Modifies code/__HELPERS/game.dm


Rohesie:
Tested without issues, @ninjanomnom

ninjanomnom:
You need to profile this sort of change. Also the old code isnt recursion, `get_turf()` isnt `get_area()`

PR #389, opened 2018-09-27
Shotguns Equipable to Back

Surrealaser:
Per title.

No more needing a sheath to do what rifles already can.


Rohesie:
The sprite only accounts for the M37.

Surrealaser:
@Rohesie Close enough for half of them.

PR #395, opened 2018-09-29
Big Xeno Rebalance

Rohesie:
* Experimental rework to quicken roundstarts:
1. Dropship can be launched10 minutes into the round, instead of 25.
2. Fog lifts at 30 minutes plus random time, instead of 45.
3. Evolution and upgrade timers cut down for everyone but the queen. The genera formula is (100 x age x tier) ticks. Young is age 1, T1s are tier 1, and queens are considered tier 4 here. Carrier and Hivelords down to T2 timers, as is their caste, drones back to T1 as well.
* Boiler isn't zoomed out on click. Just a little QoL change, no impact on balance.
* Queen death disrupts the hivemind for 2.5 minutes, after which xenos can use it again. It's half the time needed for a new queen to evolve.
* Warriors slower and tankier. Agility gives a speed boost at the cost of reducing armor. Max speed warrior is still slower than before. Small nerf to the lunge and grab, making it more in line with the Lurker's pounce (though still arguably stronger).
* Defenders less armored but healthier. Crest defense and fortify buffed.
* Being nested increases larva growth in a proportion of +3% per health point, instead of the flat +150% (or +50% pre-buff).
* Facehugger stun reduced from 100 to 60 seconds.
* Nest resist time reduced from 120 to 80 seconds.
* Flatten the aging power-gain curve. T1s start squishy but at Mature they get a huge boost, ready for battle. Elite and Ancient give minor benefits. T2s and T3s start strong and aging gives them little benefits. No T2s and T3s are as weak as their former Young counterparts, nor as strong as their former Ancient counterparts.
* Xenos slowed down while dragging humans, dead or alive. Drones, sentinels and defenders quite buffed in their general drag speed (good to drag monkeys at roundstart, or deal with revolting hosts in the hive). Runners lose their bonus there (to reduce the marine-kidnap meta, they can still devour monkeys/hosts without fear of harming them).
* Warding pheromones increase xeno armor a little. It was pre-coded but not fully implemented.
* Fixed bugs related to retaining speed and armor bonuses or penalties while aging if in agility, crest defense or fortify modes.
* Join-as-larva verb added for ghosts and new players in the lobby. Now xeno candidates can join as a burrowed larva whenever they meet the conditions instead of being forced to wait for RNG to spawn them.
* Due to popular demand, as voted in the discord poll, Lurker is named back to Hunter and Elites become Elders.
* All builders of xeno structures do so at a speed relating to their health. One second per structure if at full health, four if at 0 health, just before they drop.
* Thick walls down from 400 to 300 health (bullets need to do 600 damage to break them). Normal walls have 200 health, for reference.


ninjanomnom:
The only ones I have minor issue with here are:
* Xenos can still communicate and see the hive status on queen death.
* Building xeno walls and doors takes much longer. A little counter to the hivelord wall spam meta.

The first I feel would reduce the interesting confusion that happens when the queen dies.

The second hurts what makes hivelords unique. The better nerf imo as far as that goal goes is a reduced build speed overall when damaged and improving the marine capability to clear structures when not in danger. I agree hivelords need some changes but they should be oriented around making them more specialized for their task and giving them interesting weaknesses.

code/game/objects/structures/stool_bed_chair_nest/xeno_nest.dm
```diff
@@ -12,7 +12,7 @@
 	var/on_fire = 0
 	var/resisting = 0
 	var/resisting_ready = 0
-	var/nest_resist_time = 1200
+	var/nest_resist_time = 800 //1.5 mins
```
ninjanomnom: We have a define for this btw `1.5 MINUTES` works because minutes is `#define MINUTES *600`

code/modules/cm_aliens/XenoStructures.dm
```diff
@@ -311,18 +314,29 @@
...
-	health -= Proj.damage/2
+	health -= Proj.damage * 0.5
 	..()
 	healthcheck()
-	return 1
+	return TRUE
+
+/obj/structure/mineral_door/resin/flamer_fire_act()
+	health -= 50
+	if(health < 0)
```
ninjanomnom: `<=`?

code/modules/mob/living/carbon/xenomorph/Castes/Boiler.dm
```diff
@@ -48,20 +48,20 @@
...
+	smoke.attach(src)
+	see_in_dark = 20
+	ammo = ammo_list[/datum/ammo/xeno/boiler_gas]
+
+/mob/living/carbon/Xenomorph/Boiler/Dispose()
+	SetLuminosity(0)
+	if(smoke)
+		cdel(smoke)
+		smoke = null
+	. = ..()
```
ninjanomnom: do `return ..()` when it's at the end

code/modules/mob/living/carbon/xenomorph/Castes/Ravager.dm
```diff
@@ -107,10 +107,10 @@
...
 		)
 
-	New()
-		..()
-		verbs -= /mob/living/carbon/Xenomorph/verb/hive_status
-		spawn(15) name = "Ravenger"
+/mob/living/carbon/Xenomorph/Ravager/ravenger/New()
+	..()
+	verbs -= /mob/living/carbon/Xenomorph/verb/hive_status
+	spawn(15) name = "Ravenger"
```
ninjanomnom: ...Why the hell is this spawned

Rohesie:
1) Xenos not being able to communicate makes for a poor no-queen roundstart, and seriously disrupts them. The main issue is them being unable to communicate on who'll be the new queen, with the possibility of players wanting to be the queen and others not wanting feeling forced to. Also punishes the whole hive for the queen's mistake.
What do you suggest instead? A short hivemind blackout on queen's death, perhaps??

2) The nerf is not huge, and hivelords can still spam every other xeno structure, or reinforce existing ones without delay. They are more buffed as of now. Perhaps lowering their tunnel construction cooldown instead?
Could also add a wounded-slowdown to building, but it wouldn't solve issues like this:
yzryzew

code/_onclick/xeno.dm
```diff
@@ -30,19 +30,14 @@
...
-				bomb_turf(A)
-			else if(isturf(get_turf(A)))
-				bomb_turf(get_turf(A))
-			if(client)
-				client.mouse_pointer_icon = initial(client.mouse_pointer_icon)
-			return 1
+	if(!istype(A,/obj/screen) && is_bombarding)
+		if(isturf(A))
+			bomb_turf(A)
+		else if(isturf(get_turf(A)))
```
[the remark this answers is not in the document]
ninjanomnom: `get_turf()` always returns a turf or null

Rohesie:
* Queen death disrupts the hivemind for 2.5 minutes, after which xenos can use it again. It's half the time needed for a new queen to evolve.
* All builders of xeno structures do so at a speed relating to their health. One second per structure if at full health, four if at 0 health, just before they drop. Hivelords can still instantly upgrade existing structures without delay.
* Thick walls down from 400 to 300 health (bullets need to do 600 damage to break them). Normal walls have 200 health, for reference.

Want things tweaked further, @ninjanomnom ?

code/game/gamemodes/cm_initialize.dm
```diff
@@ -353,16 +353,6 @@ datum/game_mode/proc/initialize_special_clamps()
...
-	if(!xeno_candidate || !xeno_candidate.client)
-		return FALSE
-	if(!ticker.mode.stored_larva)
-		to_chat(xeno_candidate, "<span class='warning'>There are no longer burrowed larvas available.</span>")
-		return FALSE
-	if(!mother.ovipositor || mother.is_mob_incapacitated(TRUE))
-		to_chat(xeno_candidate, "<span class='warning'>This mother is no longer in a state to receive us.</span>")
-		return FALSE
 	else
 		return mother
```
ninjanomnom: Prety sure you don't want this under an else now

code/game/gamemodes/cm_initialize.dm
```diff
@@ -406,7 +396,7 @@ datum/game_mode/proc/initialize_special_clamps()
 		return pick(available_xenos_non_ssd) //Just picks something at random.
 
 	new_xeno = input("Available Xenomorphs") as null|anything in available_xenos
-	if (!istype(new_xeno) || !xeno_candidate)
+	if (!istype(new_xeno) || !xeno_candidate || !xeno_candidate.client)
```
ninjanomnom: 512 you can further combine things like this and it improves performance very slightly to use
`if(!xeno_candidate?.client)`

code/modules/mob/living/carbon/xenomorph/Castes/Ravager.dm
```diff
@@ -110,7 +110,7 @@
 /mob/living/carbon/Xenomorph/Ravager/ravenger/New()
 	..()
 	verbs -= /mob/living/carbon/Xenomorph/verb/hive_status
-	spawn(15) name = "Ravenger"
+	name = "Ravenger"
```
ninjanomnom: No I was asking why this was spawned before. This looks like some fix that was added in cm because of some race condition bug. You shouldn't need this at all since the name is set in the type definition.

code/game/gamemodes/cm_initialize.dm
Rohesie: Very useful. Lots of instances that could use this kind of thing. I'll start using it from now on.

PR #399, opened 2018-09-29
Almayer Map Resize

Alterist:
Makes the Almayer map 255 x 255, hopefully reducing load times and lag
Small fixes that shouldn't affect anything


Rohesie:
This was tested, though not profiled.

PR #397, opened 2018-09-29
Adds ammo boxes into the game.

DominikPanic:
Custom sprites, ammo boxes of all rifle and SMG round types plus slug and buckshot.

PR #400, opened 2018-09-29
Req vendors rebalance and cleanup.

DominikPanic:
Removes useless items. Removes items that start empty. Adds ammo boxes. Tweaks the numbers a bit so it all makes more sense. Also tweaks the contraband.


Supermichael777:
/obj/item/ammo_magazine/shotgunbox doesn't exist.

Everything in the vendors is there for a reason, most of the empty units represent things that can be reloaded with the restock feature (unless they removed that)

Additionally a lot of the stuff in your list dosen't appear to exist.

Remember this list is not verified by the compiler. only a manual check will id flaws in the list.

subject217:
DNM without https://github.com/ColonialMarines-Mirror/ColonialMarines-2018/pull/397

Rohesie:
Some of those things can go, but not all. Remember they can still be used for admins to make events if they stay.

That or make a new vendor for them to do that.

code/game/machinery/vending/marine_vending.dm
```diff
@@ -84,53 +84,53 @@
...
-					/obj/item/storage/box/nade_box = 4,
-					/obj/item/explosive/grenade/frag = 2,
+					/obj/item/storage/pouch/magazine/pistol/large = 5,
+					/obj/item/weapon/gun/pistol/m4a3 = 20,
+					/obj/item/weapon/gun/pistol/m1911 = 2,
+					/obj/item/weapon/gun/revolver/m44 = 10,
+					/obj/item/weapon/gun/smg/m39 = 15,
+					/obj/item/weapon/gun/rifle/m41a = 20,
+					/obj/item/weapon/gun/shotgun/pump = 10,
+					/obj/item/weapon/gun/shotgun/combat = 1,
```
Rohesie: Don't add these before we nerf them.

code/game/machinery/vending/marine_vending.dm
```diff
@@ -84,53 +84,53 @@
...
+					/obj/item/explosive/mine = 2,
+					/obj/item/storage/box/nade_box = 2,
 					/obj/item/explosive/grenade/frag/m15 = 2,
 					/obj/item/explosive/grenade/incendiary = 4,
-					/obj/item/explosive/grenade/smokebomb = 2,
-					/obj/item/explosive/grenade/phosphorus = 0,
-					/obj/item/storage/box/m94 = 8,
-					/obj/item/storage/box/zipcuffs = 4,
+					/obj/item/explosive/grenade/smokebomb = 5,
```
Rohesie: Add the new tactical smoke bombs so we can field test them.

DominikPanic:
Alright, I'll list all changes now:

Moved every item that started empty to the bottom of the contraband list.
Removed a lot of the ammo from the vendors to compensate for the newly added ammo boxes.
Added ammo boxes into the supply console.
Made several useless items that were NEVER ordered in the console cheaper, slightly buffed MRE crate.
Added cloak nades as requested.
Slightly tweaked numbers of a few items, should have minimal impact, just did some rounding that made sense. If all the guns are melted, it should give about a stack of metal which shouldn't be an issue.

PR #402, opened 2018-09-29
Powerloader Resprite and Rework

Alterist:
Makes the powerloader less lawsuit-liable, removes WY reference in examine text


Alterist:
scrnshot1

Rohesie:
Sprite and name are great. Not so sure on the health and damage boosts.
The ability to carry more crates than just two would be really good, though.

PR #403, opened 2018-09-29
Mentor powercreep pr 1: Mentors can now hand out the BEST COOKIE

F-OS:
Pros: Cookie

Cons: Idk can someone with more experience tell me if this might be a problem?


factoryman942:
help mentors cookiebusing me

subject217:
Mentors don't really have any guidelines on what they're doing so letting them interact in any way that isn't simply an OOC dialogue is suspect IMO. Would it be a problem? Probably not. Could it be a problem? Potentially.

Rohesie:
I personally don't mind, but not sure it fits the vision of @francinum so it's up to him.

ninjanomnom:
Not a fan of mentors being able to have any tangible physical effect on the round. There is no reason they need to have this capability

PR #405, opened 2018-09-30
makes movement responsive

kachnov:


Rohesie:
Been tested. The difference is noticeable.

PR #406, opened 2018-09-30
Delets all donor items sprites and related files

Alterist:
DON'T MIND ME, JUST TAKING OUT THE TRASH


kachnov:
ok now this is epic

MrStonedOne:
How about make them byond member only.

Alterist:
I had some thinking about this, while I would like to keep them for events and whatnot, there are 3 solid reason why they need to be gone
1. They're a nightmare of copyrights and IPs, from MGS to Star Wars to Fallout to Doom to Metroids!
2. These are CM's donator rewards which are sets of marine armor that has special apperances,  in which case they might be considered artworks paid for by individual donors. 
3. We are supposed to be getting rid of CMs unique sprites as a sign of goodwill, even if there might not be an legal impetus to do so

francinum:
Take the icons, leave `code/modules/cm_marines/Donor_Items.dm`

Rohesie:
Why leave that file?

I like Alterist's reasoning. As much as I enjoy some of those sprites they should go, out of respect.

Alterist:
After more digging I realized two things:
1. That the icons for all of the donor items are actually inside the icons\mob folders, so more deletions incoming.
2. Some of the donor sprites were actually just sprites from tg's codebase.
This really don't sit right with me, It's one thing to take an open codebase and make it closed source, it's another to take the art from it and give it out as an donation reward, so here's a compromise: any icon that has the name of the donor will be removed, as is any icon that might attract the attention of certain companies (one of them starts with N and the other starts with D)

Alterist:
suit
mask
feet
head
uniform
An overview of the deleted sprites

PR #418, opened 2018-10-03
Medkits and Auto-equip

Rohesie:
- Medkits are large and don't fit in backpacks anymore, to avoid the magical multiplication of storage space. In compensation they provide much more useful contents.
- Doctors start with a surgical mask, a medkit pouch instead of the useless syringe pouch, and a medbelt which contains a stasis bag, roller bed and defib. They should be ready to heal, minimizing prep time.
- MTs start with welding goggles and a welding pack in their hand instead of the useless scanner.
- Auto-equip functions implemented. Press E to put items inside an open container or inside a pouch. Like the medkit one, for example.
- Knives can be put in boots with E, and ties (webbing/vest) in jumpsuits.


Surrealaser:
@Rohesie Regular backpacks should be able to fit medikits; might make them worth at least half a shit.

code/game/objects/items.dm
```diff
@@ -479,6 +479,24 @@ cases. Override_icon_state should be a list.*/
 					if(S.can_be_inserted(src))
 						return TRUE
 				return FALSE
+			if(EQUIP_IN_STORAGE)
+				if(H.s_active)
+					var/obj/item/storage/S = H.s_active
+					if(S.can_be_inserted(src))
+						return TRUE
+				return FALSE
```
Qustinnus: `if (!H.s_active)
    return FALSE
`

avoid unnecesary nesting

code/game/objects/items.dm
```diff
@@ -479,6 +479,24 @@ cases. Override_icon_state should be a list.*/
...
+			if(EQUIP_IN_L_POUCH)
+				if(H.l_store && istype(H.l_store, /obj/item/storage/pouch))
+					var/obj/item/storage/S = H.l_store
+					if(S.can_be_inserted(src))
+						return TRUE
+				return FALSE
```
Qustinnus: ditto

code/game/objects/items.dm
```diff
@@ -479,6 +479,24 @@ cases. Override_icon_state should be a list.*/
...
+			if(EQUIP_IN_R_POUCH)
+				if(H.r_store && istype(H.r_store, /obj/item/storage/pouch))
+					var/obj/item/storage/S = H.r_store
+					if(S.can_be_inserted(src))
+						return TRUE
+				return FALSE
```
Qustinnus: ditto

code/modules/mob/mob.dm
```diff
@@ -138,13 +138,16 @@
...
-	if(!istype(W)) return
+/mob/proc/equip_to_slot_if_possible(obj/item/W, slot, ignore_delay = TRUE, del_on_fail = FALSE, disable_warning = FALSE, redraw_mob = TRUE, permanent = FALSE)
+	if(!istype(W))
+		return
 
 	if(!W.mob_can_equip(src, slot, disable_warning))
-		if(del_on_fail) cdel(W)
+		if(del_on_fail)
+			cdel(W)
 		else
```
Qustinnus: else if

Supermichael777:
Questnnus that's just how TG roles, both are equivalent anyhow

Qustinnus:
they are not equivelant.

Rohesie:
This has been testmerged for a while.

code/modules/mob/living/carbon/human/inventory.dm
```diff
@@ -71,18 +71,26 @@
 			return has_limb("chest")
 		if(WEAR_ACCESSORY)
 			return has_limb("chest")
+		if(EQUIP_IN_BOOT)
+			return has_limb("r_foot") && has_limb("l_foot")
 		if(WEAR_IN_BACK)
-			return TRUE
+			return has_limb("chest")
```
Qustinnus: im pretty sure you cant remove your chest

Rohesie: I'm just replicating already existing checks. The purpose I can see is for re-using the proc with modified species that descend from humans (or different if the proc is changed to `living` or `mob`.

Qustinnus: I suppose, I feel like it's a bit of a waste of cpu use but I suppose there is always modularity to think of.

PR #422, opened 2018-10-04
Set Respawn Timer admin verb.

DominikPanic:
Temporary lowpop bandaid until we get a better solution. It's still an admin-only verb that starts disabled so it shouldn't cause any issues at the moment.

### DIFF
--- a/code/modules/admin/admin.dm
+++ b/code/modules/admin/admin.dm
@@ -1,6 +1,6 @@
-
 var/global/BSACooldown = 0
 var/global/floorIsLava = 0
+var/global/respawntime = 15
 
 
 ////////////////////////////////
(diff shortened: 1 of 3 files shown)


Rohesie:
Make it a variable timer you can edit as an admin verb instead.

Surrealaser:
You should scale the reduction in proportion to the population under a target level, rather than just straight up halving it.

DominikPanic:
Alright so there should be a new admin verb that sets the respawn timer in minutes. Default still 15.

code/modules/admin/admin.dm
```diff
@@ -849,6 +849,17 @@ var/global/floorIsLava = 0
 	world.update_status()
 	feedback_add_details("admin_verb","TR") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
 
+/datum/admins/proc/toggleatime()
+	set category = "Server"
+	set desc="Sets the respawn time"
+	set name="Set Respawn Timer"
+	var/message = input("Enter time in minutes:") as message
+	respawntime = text2num(message)
```
MrStonedOne: this should be an arg to the proc, not a var.

This way the client is handles showing them the prompt when they click, rather than requiring a round trip to the server.

```dm
/datum/admins/proc/toggleatime(respawntime as num)
```
Byond will handle sanitizing it as a number.

DominikPanic:
Should be sanitized properly now.

code/modules/admin/admin.dm
```diff
@@ -853,7 +853,10 @@ var/global/respawntime = 15
 	set category = "Server"
 	set desc="Sets the respawn time"
 	set name="Set Respawn Timer"
-	respawntime = time
+	if (time > 0)
```
Rohesie: Larger or equal. If for some reason they want players not have to wait at all, that should be a possibility.

PR #423, opened 2018-10-05
Minor number correction for ancient carrier description

CosmicScientist:
I don't know if "blah [CA.huggers_max] blah" would work since I don't know how references are stored in strings and I'm not prepared to set up a test host to find out for a one digit change.


Ghommie:
No need to clutter the desc, normally those values always stay the same unless varedited. Plus it's more of a brag flavor text than else.

Qustinnus:
[CA.max_huggers] would work but youd have to change max huggers before changing the desc

PR #424, opened 2018-10-05
Improved update_transform() and lay_down()

Ghommie:
Resting has been improved, more responsive and has a small flopping down animation for hominids/primates with the improved update_transform(). And a relatively small do_after when trying to get up (still more reliable than waiting the next life cycle kicking in). Throughfully tested.
Also killed off some human/xeno specific var for badminnery resizing, replaced them with an universal one for all living mobs. Should technically trigger on the cycle's update_transform() after the var edit.


code/modules/mob/living/living_verbs.dm
```diff
@@ -275,12 +275,22 @@
...
 		return
 
-	resting = !resting
-	to_chat(src, "\blue You are now [resting ? "resting" : "getting up"]")
+	if(!resting)
+		resting = TRUE
+		to_chat(src, "<span class='notice'>You are now resting.</span>")
+		update_canmove()
+	else
+		if(!action_busy) // do_after is unoptima
```
Rohesie: Do an early return to save on indentation.

Rohesie:
Wondering if `lying` shouldn't be changed to `position_angle`, `lying_angle` or something like that, since it's no longer a boolean by this new definition.

Ghommie:
The var's solely used for lying and as a mob status, it's not meant to for any other angulation trickery as it only affect humans and monkeys' sprites, and would be perplexing having to check for `lying_angle` instead of `lying` on an if proc.
So, shortly, no need to rename it.

PR #428, opened 2018-10-06
New Runner Ability: Savage - NEEDS REVIEW

Surrealaser:
-Runners may toggle on/off the ability to Savage.

-While toggled on, and while Savage is off cooldown, Runners will make a free attack whenever they pounce someone which will deal bonus damage equal to 15 or one fifth of their remaining plasma, whichever is less. They will then lose plasma equal to five times this amount.

-Savage has a 30 second cooldown that triggers after a successful usage.

-Anything that would normally prevent ability use like the stagger status effect or predator traps will prevent Savage from working.

-The objective of this new ability is to give the Runner the capacity to engage in the hit and run/harassing playstyle it was intended to excel at.


code/modules/mob/living/carbon/xenomorph/Abilities.dm
```diff
@@ -1177,6 +1177,33 @@
...
+		X.savage = FALSE
+		to_chat(X, "<span class='xenowarning'>You untense your muscles, and relax. You will no longer savage when pouncing.</span>")
+		button.overlays.Cut()
+		button.overlays += image('icons/mob/actions.dmi', button, "savage_off")
+	else
+		X.savage = TRUE
+		var/warning = ""
+		if(X.savage_used)
+			warning = " However, you're not quite yet able to savage again."
+		to_chat(X, "<span class='xenowarning'>You ready yourself for a killing strike. You will savage when pouncing.[warning]</span>")
```
Qustinnus: to_chat(X, "<span class='xenowarning'>You ready yourself for a killing strike. You will savage when pouncing.[X.savaged_used ? " However, you're not quite yet able to savage again." : ""]</span>") might be better instead of the var/warning stuff

code/modules/mob/living/carbon/xenomorph/Powers.dm
```diff
@@ -1436,3 +1436,40 @@
...
+		return
+
+	src.visible_message("<span class='danger'>\ [src] savages [M]!</span>", \
+	"<span class='xenodanger'>You savage [M]!</span>", null, 5)
+	var/extra_dam = min(35, plasma_stored * 0.25)
+	M.attack_alien(src,  extra_dam) //Inflict a free attack on pounce that deals +1 extra damage per 4 plasma stored, up to 35 or twice the max damage of an Ancient Runner attack.
+	use_plasma(extra_dam * 4) //Expend plasma equal to 4 times the extra damage.
+	do_savage_cooldown()
+
+	return 1
```
Qustinnus: return TRUE

Surrealaser:
@Rohesie 

This is why it's on a 30 second cooldown, but sure, I can add a small baseline cost.

Surrealaser:
Why is this build taking forever to process?

Qustinnus:
It finished but for some reason didn't update here

Rohesie:
Have you tested this locally?

Surrealaser:
@Rohesie Just did

Surrealaser:
Not sure why this build is stalling out; I literally just compiled it without problems.

PR #430, opened 2018-10-06
Fixes xeno stun due to knocked_out not being decremented

CosmicScientist:
One line change in the xenomorph handle_statuses() proc, added handled_knocked_out() which was reserved for use by simple_animal because every other mob handles it differently... for some reason I unno I haven't been interested enough to look in the files.

The new behaviour should decrement knocked_out by 1 each time handle_statuses is called.

A more elegant solution would be to rewrite all mobs to all call super/..() on their handle_statuses() and modify their version of the procs called but I'm giving this bugfix before considering to undertake what might as well be a rewrite of all mob code for how get/set/modify functions are and aren't used for changing vars.


CosmicScientist:
Sorry, probably not being entirely clear.

Only xenomorphs appeared to not decrement knocked_out, which resulted in a permanent stun whenever a xenomorph had knocked_out > 0, so I've called a function which decrements it to fix that.

Ghommie:
Yea, I have already coded the "elegant" solution in on my Chem/Life PR, no worries using the hacky one for now.

But I got to agree with whoever commented in code that life() is a rollercoaster of f-stupid.

PR #433, opened 2018-10-07
All the standard admin movement verbs no longer log when said admin is observing.

F-OS:
Almost all of these are standard observer commands and logging them for ghosted admins is unneeded.


Rohesie:
Hmm, I'd like some feedback from admins here.

subject217:
This is fine. As a follow up, almost all those JMP buttons should be replaced with FLW when they apply to mobs, which doesn't print to chat anyway and isn't usable as a living mob.

PR #437, opened 2018-10-08
Makes ap rockets actualy do direct hit damage

Supermichael777:
Say it with me: Don't make sub-types that override all the parents unique behaviors.  Especially if that parent overrode something you needed.
🆑
fix: AP rockets actually do impact damage.
/:cl:

Fixes some high quality cm code, that was clearly labeled but never actually fixed by anyone.
This probably makes ap rockets too good.

### DIFF
--- a/code/modules/projectiles/updated_projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/updated_projectiles/ammo_datums.dm
@@ -848,38 +848,49 @@
 	smoke.set_up(1, get_turf(P))
 	smoke.start()
 
-/datum/ammo/rocket/ap
+/datum/ammo/rocketap
 	name = "anti-armor rocket"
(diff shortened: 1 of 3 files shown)


code/modules/projectiles/updated_projectiles/ammo_datums.dm
```diff
@@ -848,38 +848,49 @@
 	smoke.set_up(1, get_turf(P))
 	smoke.start()
 
-/datum/ammo/rocket/ap
+/datum/ammo/rocketap
```
Qustinnus: this is bad OOP

Qustinnus: make rocket/normal and rocket/ap not rocket and rocketap

PR #439, opened 2018-10-09
Synth Fix

Surrealaser:
-Synths can no longer directly activate grenades as it's against their programming.

-Telescopic baton damage reduced to at least semi-reasonable levels.

### DIFF
--- a/code/game/objects/items/weapons/swords_axes_etc.dm
+++ b/code/game/objects/items/weapons/swords_axes_etc.dm
@@ -70,7 +70,7 @@
 		icon_state = "telebaton_1"
 		item_state = "telebaton_1"
 		w_class = 3
-		force = 40//Super Robust - Probably shouldn't give it to anyone who isn't staff unless you want BEATINGS
+		force = 10//Super Robust - Probably shouldn't give it to anyone who isn't staff unless you want BEATINGS
 		attack_verb = list("smacked", "struck", "slapped")
 	else
 		user.visible_message("\blue [user] collapses their telescopic baton.",\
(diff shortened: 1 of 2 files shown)


code/game/objects/items/weapons/swords_axes_etc.dm
```diff
@@ -70,7 +70,7 @@
 		icon_state = "telebaton_1"
 		item_state = "telebaton_1"
 		w_class = 3
-		force = 50//Super Robust - Probably shouldn't give it to anyone who isn't staff unless you want BEATINGS
+		force = 40//Super Robust - Probably shouldn't give it to anyone who isn't staff unless you want BEATINGS
```
Rohesie: This is still not acceptable. We want the baton to be something more like /tg/'s:
https://github.com/tgstation/tgstation/blob/e5e0459e334f2c71caf64f111cdd425f2991a89d/code/game/objects/items/melee/misc.dm

Surrealaser: It would be utterly worthless at 10 damage though.

Personally I don't give a shit so sure.

code/game/objects/items/weapons/swords_axes_etc.dm
```diff
@@ -70,7 +70,7 @@
 		icon_state = "telebaton_1"
 		item_state = "telebaton_1"
 		w_class = 3
-		force = 50//Super Robust - Probably shouldn't give it to anyone who isn't staff unless you want BEATINGS
+		force = 10//Super Robust - Probably shouldn't give it to anyone who isn't staff unless you want BEATINGS
```
Rohesie: The comment should be updated/removed.

PR #440, opened 2018-10-09
Minor Attachments Rebalance

Surrealaser:
-Recoil compensator no longer increases accuracy; instead it now has a unique niche significantly reducing recoil and spread, for both wielded and unwielded weapons.

-Stocks now increase instead of decrease accuracy penalties from movement.

-Stocks provide no benefit when one handing excepting minor benefits for the revolver (because it still applies penalties even then). As compensation their wielded recoil, scatter, and/or accuracy benefits have been increased.

-Stock wield delays changed to be proportionate to weapon size and mileage gained from the stock. 

-Made laser sight improve accuracy only because I can't see why it would ever decrease scatter; as compensation its movement penalty reduction has been increased.

-Gyro has been refocused on improving one-handed shooting by making, recoil, scatter and accuracy improvements exclusive to one handing. Further, its crippling fire delay increment has been removed, bringing it in line with other high-demand attachments for this slot.

-Vertical grip now reduces recoil by slightly more when wielded (not that I see this really coming up ever).

### DIFF
--- a/code/modules/projectiles/updated_projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/updated_projectiles/gun_attachables.dm
@@ -975,24 +975,22 @@ obj/item/attachable/attack_hand(var/mob/user as mob)
 
 /obj/item/attachable/gyro
 	name = "gyroscopic stabilizer"
-	desc = "A set of weights and balances to stabilize the weapon when fired with one hand, reducing burst scatter and movement penalties to accuracy. Slightly decrease firing speed."
+	desc = "A set of weights and balances to stabilize the weapon when shooting one-handed, burst firing or moving. Significantly reduces burst scatter, movement penalties and one-handed recoil and scatter."
 	icon_state = "gyro"
 	attach_icon = "gyro_a"
 	slot = "under"
 
 /obj/item/attachable/gyro/New()
 	..()
-	delay_mod = config.low_fire_delay
-	scatter_mod = -config.min_scatter_value
 	burst_scatter_mod = -2
 	movement_acc_penalty_mod = -3
-	scatter_unwielded_mod = -config.med_scatter_value
-	accuracy_unwielded_mod = config.low_hit_accuracy_mult
-
-
+	scatter_unwielded_mod = -config.low_scatter_value
+	accuracy_unwielded_mod = config.min_hit_accuracy_mult
+	recoil_unwielded_mod = -config.low_recoil_value
+	
 /obj/item/attachable/lasersight
 	name = "laser sight"
-	desc = "A laser sight placed under the barrel. Increases accuracy, and decrease scatter when firing one-handed."
+	desc = "A laser sight placed under the barrel. Significantly increases accuracy while moving, whether wielding or one-handed."
 	icon_state = "lasersight"
 	attach_icon = "lasersight_a"
 	slot = "under"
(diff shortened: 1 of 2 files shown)


code/modules/projectiles/updated_projectiles/gun_attachables.dm
```diff
@@ -555,19 +555,16 @@ obj/item/attachable/attack_hand(var/mob/user as mob)
 	name = "\improper M37 wooden stock"
 	desc = "A non-standard heavy wooden stock for the M37 Shotgun. Less quick and more cumbersome than the standard issue stakeout, but reduces recoil and improves accuracy. Allegedly makes a pretty good club in a fight too.."
 	slot = "stock"
-	wield_delay_mod = WIELD_DELAY_NORMAL
+	wield_delay_mod = WIELD_DELAY_FAST
```
Rohesie: This means no wield delay.

code/modules/projectiles/updated_projectiles/gun_attachables.dm
```diff
@@ -651,21 +637,17 @@ obj/item/attachable/attack_hand(var/mob/user as mob)
...
+	scatter_mod = -config.med_scatter_value
+	movement_acc_penalty_mod = 1
 
 
 /obj/item/attachable/stock/revolver
 	name = "\improper M44 magnum sharpshooter stock"
 	desc = "A wooden stock modified for use on a 44-magnum. Increases accuracy and reduces recoil at the expense of handling and agility. Less effective in melee as well"
 	slot = "stock"
-	wield_delay_mod = WIELD_DELAY_FAST
+	wield_delay_mod = WIELD_DELAY_VERY_FAST
```
Rohesie: The former delay was already zero unless another delay attachment was added.

code/modules/projectiles/updated_projectiles/gun_attachables.dm
```diff
@@ -555,19 +555,16 @@ obj/item/attachable/attack_hand(var/mob/user as mob)
...
-	recoil_mod = -config.min_recoil_value
-	scatter_mod = -config.min_scatter_value
-	movement_acc_penalty_mod = -1
-	accuracy_unwielded_mod = config.min_hit_accuracy_mult
-	recoil_unwielded_mod = -config.min_recoil_value
-	scatter_unwielded_mod = -config.min_scatter_value
+	accuracy_mod = config.low_hit_accuracy_mult
+	recoil_mod = -config.med_recoil_value
+	scatter_mod = -config.med_scatter_value
+	movement_acc_penalty_mod = 1
```
Rohesie: These magic numbers could use config values.

code/modules/projectiles/updated_projectiles/gun_attachables.dm
Surrealaser: Like I said, it's additive; it is in addition to the base delay.

Look at how the angled grip works.

code/modules/projectiles/updated_projectiles/gun_attachables.dm
Surrealaser: ?

code/modules/projectiles/updated_projectiles/gun_attachables.dm
Rohesie: It's not really hard.

Surrealaser:
@Rohesie must've accidentally saved that with the other changes in the defines.

Anyways, I'd say just test merge this and close #443; I only wanted to tweak the falloff of SMGs really, and so far as I understand, the SMG burst changes are already in.

PR #442, opened 2018-10-09
M37 Pump Lock and Notification

Surrealaser:
-M37 now has a pump lock like most modern pump shotguns.

-Pump lock is released by unloading a shell or firing the weapon.

-Bold notifications are provided when trying to pump while the lock is in place, and after successfully pumping.


code/modules/projectiles/updated_projectiles/guns/shotguns.dm
```diff
@@ -65,10 +65,12 @@ can cause issues with ammo types getting mixed up during the burst.
 			playsound(user, reload_sound, 25, 1)
 			new_handful.forceMove(get_turf(src))
 		else
```
Qustinnus: else if

PR #443, opened 2018-10-09
Further SMG Improvements

Surrealaser:
-Wielding and one-handing accuracy and scatter slightly improved.

-Falloff reverted back to -1 per tile traveled.

-Gyro now works with the SMG.


Rohesie:
No other changes besides the falloff?

Surrealaser:
@Rohesie At this point, yes.

Rohesie:
Document your changes. This is not the only thing the PR modifies.

Surrealaser:
@Rohesie Those are the current SMG values in place on the test merge, or they should be.

code/modules/projectiles/updated_projectiles/guns/smgs.dm
```diff
@@ -60,8 +60,8 @@
 
 /obj/item/weapon/gun/smg/m39/set_gun_config_values()
 	fire_delay = config.low_fire_delay
-	burst_delay = config.mlow_fire_delay
-	burst_amount = config.high_burst_value
+	burst_delay = config.min_fire_delay
+	burst_amount = config.med_burst_value
```
Rohesie: This is an undocumented change.

Surrealaser: As above.

It's per the test merge; you merged the older iteration of this which had same values.

Rohesie:
Perhaps revert the falloff back to the usual 1 and worsen the gun while unwielded, since it gives no wielded slowdown. Then we can repurpose the pistols as viable one-handed weapons.

PR #446, opened 2018-10-10
Partial change of almayer area sprites

CosmicScientist:
from:
old_area_almayer
to:
area_almayer

Francinum wanted to not see a funny shape when mapping out the AI's chamber and most of the rest of the mapping area sprites are symbolic in that you can learn them if you look at them, for an hour, but for the most part aren't informative to someone just picking it up.

I've applied some symbols sparingly since CIC was something that took me a while on CM to understand it meant Combat Information Centre because Bridge is passé.

I didn't replace all the sprites because it really does take hours to put in pixels as intelligible words on a 64x64 and I'm too much of a neanderthal to discover a better way.


CosmicScientist:
I might come back at another date, with another PR, to replace more area sprites but this should make it *less* painful to remap the ship.

Supermichael777:
>once you see it
Kek

PR #448, opened 2018-10-10
Devour Tweak

Surrealaser:
-Devoured marines are now blinded for the duration of the devour, allowing aliens the opportunity to deposit them in unrecognizable locations.

-Devour delay now scales inversely with the pain the victim is in; the higher the pain (i.e. they're less able to resist), the faster it is completed, with a current max delay of 20 seconds vs a full health/no pain marine, and a minimum delay of 3 seconds. The delay is reduced by 1 second per 10 pain.

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/life.dm
+++ b/code/modules/mob/living/carbon/xenomorph/life.dm
@@ -38,7 +38,7 @@ adjustFireLoss(-(maxHealth / 70 + 0.5 + (maxHealth / 70) * recovery_aura/2)*(m))
 				if(loc != zoom_turf || lying)
 					zoom_out()
 			update_progression()
-			update_evolving()	
+			update_evolving()
 			//Status updates, death etc.
 			handle_aura_emiter()
 			handle_aura_receiver()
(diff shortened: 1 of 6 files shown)


code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -103,6 +103,9 @@ adjustFireLoss(-(maxHealth / 70 + 0.5 + (maxHealth / 70) * recovery_aura/2)*(m))
 				if(M.loc != src)
 					continue
 				M.forceMove(loc)
+				if(isliving(M))
+					var/mob/living/L = M
+					L.blinded = FALSE
```
Rohesie: Since this code is repeating in each instance, it could become a proc under `/mob/` to simplify things.
Also this can act as an unintended cure for blindness. Make the blind removal check for other conditions.

Surrealaser: Sure, I'll run a check for the blind disability.

The only thing that this might interfere with is blind stacks, but their blind gets quickly reapplied anyways if the victim still has stacks.

Surrealaser:
Closed in favour of #453

PR #453, opened 2018-10-11
Devour Tweak 2.0

Surrealaser:
-Devoured marines are now blinded for the duration of the devour, allowing aliens the opportunity to deposit them in unrecognizable locations.

-Devour delay now scales inversely with the pain the victim is in; the higher the pain (i.e. they're less able to resist), the faster it is completed, with a current max delay of 20 seconds vs a full health/no pain marine, and a minimum delay of 3 seconds. The delay is reduced by 1 second per 10 pain.


subject217:
fetish content

Rohesie:
Conflicts as expected.

code/modules/mob/mob_grab.dm
```diff
@@ -97,7 +97,7 @@
 						return */
 		X.visible_message("<span class='danger'>[X] starts to devour [pulled]!</span>", \
 		"<span class='danger'>You start to devour [pulled]!</span>", null, 5)
-		if(do_after(X, 50, FALSE, 5, BUSY_ICON_HOSTILE))
+		if(do_after(X, min(30, 200 - pulled.traumatic_shock), FALSE, 5, BUSY_ICON_HOSTILE)) //Devour delay scales inversely with victim's pain; 20 seconds max, 3 seconds min
```
Rohesie: I think you want here a `max(30, 200- pulled.traumatic_shock)`, no? `CLAMP` would be safer.

Rohesie (changes requested):

code/modules/mob/living/carbon/carbon_helpers.dm
```diff
@@ -10,3 +10,31 @@
...
+		A.forceMove(loc)
+		if(ismob(A))
+			if(forced)
+				visible_message("<span class='danger'>[A] bursts out of [src]!</span>")
+			if(isliving(A))
+				var/mob/living/L = A
+				L.devoured = FALSE //remove the devour flag
+	return TRUE
+
+/mob/living/carbon/proc/unvore_forced(var/mob/living/carbon/Xenomorph/X)
```
Rohesie: This seems to overlap with `unvore()` with a `forced` flag.

code/modules/mob/living/carbon/xenomorph/XenoProcs.dm
```diff
@@ -524,3 +513,17 @@
 		leader_aura_strength = Q.aura_strength
 		leader_current_aura = Q.current_aura
 		to_chat(src, "<span class='xenowarning'>Your pheromones have changed. The Queen has new plans for the Hive.</span>")
+
+//Bleuugh
+/mob/living/carbon/Xenomorph/proc/empty_gut()
```
Rohesie: This overlaps as well.

Rohesie (changes requested):

PR #417, opened 2018-10-03
Portin PRs about Chemistry, Reagents, Life Procs, Alcohol.

Ghommie:
Out of everything I have to say, is that in spite of PR atomization, I find myself modifying way more files than expected. Second is, I tested this stuff no less than two dozen times but I would feel sorry if this somehow degrades robots and chimps (not accounting pre existing bugs) or worse.

- Revamps chemistry (machinery and colours aside) to something less outdated, same for containers.
- More reaction types for reagents: INJECT, VAPOR, PATCH. Currently serving little purpose, but at least you won't taste injected food.
- more container types, less snowflake examine strings on containers.
- tweaks on chemicals, especially since Crit OD stacks with the OD. (Overdose otherwise unchanged, triggers and de-triggers when the threshold is crossed)
- medicine and consumable (alcohol, food, drinks) reagent subtypes.
- Reworked ultrazine addiction: It is no more a virus and lasts 38 minutes (standard) instead of 2 hours. It has an addiction threshold, but it's very very low. Probably subject to changes.
- You can now taste ingested food and reagents!
- Taste sensitivity: Default to normal for most mobs, may be different for some species (Alias synths still can't eat, TODO list)
- cigarettes & co are no more nicotine free now. Of course they taste like smoke. Plus removed some junk text and buffs.
- the marine subtype flask starts with some extra hooch. Not enough to bruise the liver, just very lighty inebriate the imbiber.
- Alcohol may moderately alleviate pain at high doses, but it is way less convenient than a proper painkiller.
- Standard MREs comes with an expanded list of stupid tastes from a .txt file, because I couldn't give proper ones to each type.
- Other slight changes to food contents.
- Revamped life() a bit, especially on vision/stat related stuff. Yell out if I removed something important or a thing isn't working properly.
- Some more mob status procs.
- Masks, glasses and masks now use an equal tint var instead of specific ones for each type. Beware, stacking more than three points worth (e.g. welding googles/helmet + gas mask) of tint layers will obscure the vision.
- Blindfolds less snowflakey.
- You can't properly examine stuff anymore when blind.
- Added in slight perm coefficient to marine armor et others.
- You can unsecure a welding tool. Won't make an (impromptu) flamethrower out of it, but can empty its contents on a beaker, or refill from one. Remember though that it only works with welding fuel.

Apologises for the atomization-less PR, tough luck I found out it was a domino cascade pretty late.

PR #454, opened 2018-10-11
Slow Stacking Removed for Staggerstun

Surrealaser:
Per title.

While a Xeno has slowdown stacks, it cannot receive additional slowdown stacks from the staggerstun effect most prominently featured on slugs.


Rohesie:
Slowdown may come from more than one source in the long run, such as pulling humans and whatever else. Because we don't want effects to directly modify the speed var, just the modifier var. This means that if a xeno chooses to lose speed, like with the defender's head crest, even of only a little, they will be immune to the potentially larger slug slowdown.

Surrealaser:
> Slowdown may come from more than one source in the long run, such as pulling humans and whatever else. Because we don't want effects to directly modify the speed var, just the modifier var. This means that if a xeno chooses to lose speed, like with the defender's head crest, even of only a little, they will be immune to the potentially larger slug slowdown.

Well the alternative is that we add a separate variable for the stagger slowdown.

Rohesie:
Dependent on #417

Surrealaser:
@Rohesie Flag might work for slowed as there is no way to alter the duration, so you can add a spawn timed removal or a world time stamp that gets compared during the life process which syncs with it, but it's crude and kludgy.

I have no plans to preclude stacking for staggered though.

Rohesie:
Let's get the opinion of @ninjanomnom or @Qustinnus here.

ninjanomnom:
We actualy have a fairly recent system in tg for managing speed sources and the like. That would be ideal if you wanted to port it. This system is acceptable though.

Rohesie:
Conflicts.

PR #315, opened 2018-09-15
Improved Item sprites for some clothes (mostly headwear).

Ghommie:
Time to slighty improve clicking for several worn items which were previously just using duplicates of their mob overlays. Bigger hitboxes, centered sprites.

PR #458, opened 2018-10-11
Fixes suit lamp displaying bug

MoXuT0:
Per title.
Lamp sprite on mob is bugged and only has one direction. That fixes it.


Ghommie:
It is quite sub-optimal for such overlays to be shared on both mob and object, but it's my fault for ignorantly deleting them on #315.

PR #176, opened 2018-08-21
Adds moths

Jared-Fogle:
image

image

image

Adds moths as a playable species because maintainers said they'd only consider species from /tg/. Pretty much nothing but sprites was ported over because CM code. Also adds a good base for any future species to be added.

Unimportant TODOs:

- [ ] Ghost sprites are a bit off.
- [ ] Examine sprite doesn't show wings.

PR #460, opened 2018-10-11
Fix duplicate constant issues.

Rohesie:
Fixes having duplicate definitions of `TARGETED_LAYER`, `INFINITY` and `OPEN`. Blame #235 and #176 and the code we inherited.

Fixes some runtime errors we also inherited due to rather curious code for the hangar elevator (used to move the tank from the tank bay). The code was trying to get the `x`, `y` and `z` variables of `HangarUpperElevator` and `HangarLowerElevator` before they were created in their landmarks. This would be used later on when the ship crashes in order to play the crashing sound of the elevator falling, which can actually simply use the location of the elevator without the need of those intermediary steps.

### DIFF
--- a/code/game/objects/structures/mineral_doors.dm
+++ b/code/game/objects/structures/mineral_doors.dm
@@ -1,5 +1,5 @@
-#define CLOSED 0
-#define OPEN 1
+#define D_CLOSED 0
+#define D_OPEN 1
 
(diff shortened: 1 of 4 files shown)


Qustinnus:
be more concise in your PR info, Im not really sure what you fixed and I dont want to figure out what you fixed on my own

Rohesie:
Fixed that, @Qustinnus

PR #470, opened 2018-10-12
Quick revamp of adminhelp to make it easier to use for both players and admins

F-OS:
New adminhelp types:
"Rule Violation / Policy Question" - Sends to admins, not mentors.
"Gamebreaking Bug" - Sends to admins, special tools for bugfixes incoming, give suggestions?
"Gameplay Question" - Sends to mentors and admins.


Qustinnus:
bugs should be reported to github not to admins because that will result in brainlets PMing me at 2am about them instead of making an issue on the gihtub

Rohesie:
We don't have a developer administrative rank yet and this would require a larger rework of the system, besides hiding the initial AHelp from those without debug permissions.

F-OS:
Yeah i'll just trash that

Rohesie:
Outdated by #580

Issue #43, opened 2018-07-31
Sprite License Issue?


callanrocks:
Like half the sprites are still just /tg/station13 vanilla but some are edited or entirely original and there's no indication of what they're covered under.

As Oranges has pointed out, this is actually a problem.

Maybe start porting sprites over?

Wilchenx:
pretty much, yes

kyrahabattoir:
To be honest it's not just the sprites, it's any art assets added by the CM staff aswell: sounds, sprites, maps.

I'd argue that any writing and flavor texts is also technically part of the "content" rather than code.

MrStonedOne:
the git history can be force pushed when needed.

Modified sprites or sprites based off of /tg/ sprites retain the cc-by-sa license. The same applies to sounds.

Wholly new assets could be an issue. It is the current stance of /tg/ that we have a good faith reason to believe that our source for the repo had the require agency to release the repo to us. If anybody who contributed an art asset to CM would like to contest that, they are suggest to write to mso@tgstation13.org. Good faith attempts to resolve such issues will be respected, Opening with a dmca to github or our ISP before attempting to communicate with us will not be considered a good faith attempt, and will be fought in court.

I should also remain anybody reading this that /tg/station13 Has a Registered DMCA Agent with the United States Copyright Office, and sending a dmca to any other entity may be in violation of the Digital Millennium Copyright Act (Ask your lawyer).

While maps may be seen as assets, in CM it looks like they are compiled in (by the map daemon), making them covered under gpl.

The way the compiler reads map files follows the same parsing rules as all DM code.

That being said, anybody who is the sole or primary map creator of an original CM map may contact me at the above email address to discuss removal.

kyrahabattoir:
> Good faith attempts to resolve such issues will be respected, Opening with a dmca to github or our ISP before attempting to communicate with us will not be considered a good faith attempt, and will be fought in court.

kyrahabattoir:
https://www.reddit.com/r/SS13/comments/94rz25/an_open_reply_to_mso_please_dont_use_the_work_i/

You got your discussion here.

PR #472, opened 2018-10-12
Replace Big Red CM turf sprites with TG turf sprites

Arachnidnexus:
This PR is replacing the CM turf tiles for Big Red with sprites modified and adapted from turf sprites from tgstation. It should help with one of the many parts of #43 that need to be addressed.

CM sprites:
bigred

Modified TG sprites:
tgredv9

Dream Maker view:
image
image
image

The cave tiles could use a better randomizer, but that's probably suited for a separate PR.


Arachnidnexus:
Edited some of the transition tiles slightly to remove glaring repeats so dirt-cave should look nicer. There are still some very sharp edges between tile types though.

image

Qustinnus:
image
bays rocks look better, can we use these with a different color on them? they dont fit TGs lavaland but they would fit fine here

Arachnidnexus:
@Qustinnus I compared the Bay and CM rock sprites (walls.dmi in both cases) and the rock sprite is identical. The color value used on the rock sprite for Big Red makes it lose a bit of detail because the darker shades blend in more with the color filter.
image

PR #475, opened 2018-10-13
Extended explosion parameters, more work required

CosmicScientist:
Trying to make explosion logging better, at least so you can search who or what in the logs.

- [x] add parameters to both explosion procs
- [ ] go through all explosion() calls to try and find a reason for as many as I can
- [ ] see if adding a list of mobs affected would be painful

### DIFF
--- a/code/game/objects/explosion_recursive.dm
+++ b/code/game/objects/explosion_recursive.dm
@@ -10,7 +10,7 @@
 var/list/explosion_turfs = list()
 var/explosion_in_progress = 0
 
-proc/explosion_rec(turf/epicenter, power, verbose cause = "no explanation", explosion_source = "unknown")
+proc/explosion_rec(turf/epicenter, power, verbose_cause = "no explanation", explosion_source = "unknown")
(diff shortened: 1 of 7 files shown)


CosmicScientist:
Oh dear, I wonder why travis failed.

This is a work-in-progress.

code/game/objects/explosion_recursive.dm
```diff
@@ -10,7 +10,7 @@
 var/list/explosion_turfs = list()
 var/explosion_in_progress = 0
 
-proc/explosion_rec(turf/epicenter, power)
+proc/explosion_rec(turf/epicenter, power, verbose cause = "no explanation", explosion_source = "unknown")
```
CosmicScientist: Found why travis failed, no underscore in the two word var, will fix once on my next run.

CosmicScientist:
A few down, forty or sixty more to go.

CosmicScientist:
I'm sorry but I just don't have the energy to change all explosion calls.

PR #477, opened 2018-10-13
Fixes MREs

Alterist:
Adds a Important part of the marine diet which was previously missing to MREs
untitled-1


Alterist:
We need to have equal representation of the dietary requirements of all squads, favoring one squad over another is a massive imbalance, and should be something we strive for as we continue to create tasty food items for marines.

Rohesie:
Look, we aim for a 75% delta liver intoxication rate because imperfect balance is part of our game design.

That said, crayons of each squad color would be nicer.

Alterist:
So, what youre saying is....MOAR CRAYONS...

Rohesie:
Always.

Qustinnus:
fix the conflicts and its good

PR #478, opened 2018-10-13
Shotgun Pump Lock Tweak - No Longer Locks While Empty

Surrealaser:
Per title.


Qustinnus:
@rohesie is this worthy of a speedmerge

PR #481, opened 2018-10-13
buffs dancing

factoryman942:
Xenos can now move while using *dance


Rohesie:
Yes, that is a balance concern. I intend to remove moonwalking already.

Arachnidnexus:
Could prevent the crest randomness by dropping armor when dancing if that's a concern.

Ghommie:
well, the directional crest shielding thing is located at projectile.dm, line 604.
A simple solution would be making the conditional on a var, said var set on TRUE on the two castes, and have it set on FALSE for the duration of the dance.

Rohesie:
Seems a little convoluted for a meme, don't you think?

factoryman942:
jesus christ this was a two line change I didn't want to have to actually do stuff

factoryman942:
Just because I didn't want to doesn't mean I won't now
although it does require me knowing what to do about it

factoryman942:
that hurts @Rohesie

PR #482, opened 2018-10-13
Turns chat, github and patreon links into configs

Rohesie:
Makes it easier for downstreams to set up their own without needing to change the code.
Cleans code a little.
Standard configs are DMCA. 
Example configs are /tg/, no longer CM.

### DIFF
--- a/code/game/world.dm
+++ b/code/game/world.dm
@@ -255,8 +255,8 @@ var/world_topic_spam_protect_time = world.timeofday
 	var/s = ""
 
 	if (config && config.server_name)
-		s += "<a href=\"[config.forumurl]\"><b>[config.server_name] &#8212; [MAIN_SHIP_NAME]</b>"
-		s += "<br><img src=\"[config.forumurl]/byond_hub_logo.jpg\"></a>"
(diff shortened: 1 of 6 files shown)


code/game/world.dm
```diff
@@ -255,12 +255,25 @@ var/world_topic_spam_protect_time = world.timeofday
...
-		// s += "<a href=\"http://goo.gl/04C5lP\">Wiki</a>|<a href=\"http://goo.gl/hMmIKu\">Rules</a>"
+		if(config.chaturl)
+			s += "<a href=\"[config.chaturl]\"><b>[config.server_name] &#8212; [MAIN_SHIP_NAME]</a></b>"
+		else
+			s += "<b>[config.server_name] &#8212; [MAIN_SHIP_NAME]</b>"
 		if(ticker)
 			if(master_mode)
-				s += "<br>Map: <b>[map_tag]</b>"
+				switch(map_tag)
+					if("Ice Colony")
```
ninjanomnom: You dont need to do this here but we really need to move this sort of thing into map configs

PR #485, opened 2018-10-13
Brings centcom slightly up to date

francinum:
Also fixes the big ERT shuttle, it's still not perfect, but it's transit actually works.
Image:


Alterist:
Neat, might want to remove those mech suits just in case, considering that this area was never meant to be accessed in the first place

Rohesie:
The ERT shuttles will have to either have their own bases or be cut off from CentComm. We have all sorts of ERTs spawning there.

config/admins.txt
```diff
@@ -75,3 +75,4 @@ virtualjohn - Admin
 wjohnston - Admin
 wubli - Admin
 wilchen - Admin
+francinum - Admin
```
Rohesie: Care to put this in alphabetical order?

code/game/machinery/doors/shutters.dm
```diff
@@ -86,6 +86,12 @@
 	ex_act(severity) //immune to explosions
 		return
 
+/obj/machinery/door/poddoor/shutters/almayer/open
+	density = 0
+	opacity = 0
+	unacidable = 1
```
Rohesie: `TRUE`, `FALSE` and all that.

PR #487, opened 2018-10-14
Replacing Almayer CM walls with Baystation ones

Arachnidnexus:
The goal here is to replace the CM walls with Bay walls with similar perspective. So far all that's been accomplished is stitching together a generic wall which hopefully can be adapted in the future for all the CM walls. The only file changed so far is 'almayer.dmi' and I am making this PR to see if someone who can run the server can see how it looks and spot any glaring issues.

_Note that because I pushed the Big Red changes in #472 to my fork's master like a fool, these are included here too._

CM: 
image
Bay (stitched together manually):
dmca_bay_base


Arachnidnexus:
Haha ignore this, will reopen once I do not edit the wrong file completely.

Qustinnus:
can you remove the bigred file changes? else change the title + body to represent this and close the old bigred pr

PR #488, opened 2018-10-14
Motion Detector Hotfix

Surrealaser:
-Makes the motion detector much less resource intensive by limiting its process to things in range.


code/game/objects/items/devices/motion_detector.dm
```diff
@@ -86,7 +86,7 @@
 		human_user = loc
 
 	var/detected
-	for(var/mob/M in living_mob_list)
+	for(var/mob/living/M in range(detector_range, src))
```
Qustinnus: I dont think this is more performant, this gets a larger list than the one above most likely and has to do an expensive operation to get said list

code/game/objects/items/devices/motion_detector.dm
```diff
@@ -86,7 +86,7 @@
...
 		if(loc == null || M == null) continue
 		if(loc.z != M.z) continue
 		if(get_dist(M, src) > detector_range) continue
```
Qustinnus: also all of the above checks seem useless because range implies all of these are false

MrStonedOne: Remove this check as its no longer needed.

Qustinnus: copycat go away

code/game/objects/items/devices/motion_detector.dm
Surrealaser: I can't possibly see this being true except in extreme low pop.

Qustinnus: if it was living_players then it would definitely be true, you'd be surprised how many objects you can find in a 5x5 range of something.

Surrealaser: Apparently MSO did a performance test of this fix.

MrStonedOne:
```
/obj/item/device/motiondetector/proc/process_old                                                 0.011        0.089        0.092          375
/obj/item/device/motiondetector/proc/process_new                                                 0.004        0.054        0.054          381
```

Rohesie: https://i.gyazo.com/7a08ea4db2822acb2a042f0037e4072e.png
It does depend on the number of objects near.

code/game/objects/items/devices/motion_detector.dm
```diff
@@ -81,23 +90,27 @@
...
-		human_user = loc
-
 	var/detected
-	for(var/mob/living/M in orange(detector_range, src))
-		if(M == loc) continue //device user isn't detected
-		if(!isturf(M.loc)) continue
-		if(world.time > M.l_move_time + 20) continue //hasn't moved recently
-		if(isrobot(M)) continue
+	for(var/mob/living/M in orange(detector_range, human_user))
+		to_chat(world, "<span class='danger'>MOTION SENSOR DEBUG: Moving mob detected: [M.name]. Detector range: [detector_range].</span>")
```
Rohesie: Please don't push debug messages to PRs, especially ones that are being testmerged.

code/game/objects/items/devices/motion_detector.dm
```diff
@@ -81,31 +90,33 @@
...
+		if(world.time > M.l_move_time + 20)
+			to_chat(world, "<span class='danger'>MOTION SENSOR DEBUG: Aborted: Subject hasn't moved.</span>")
+			continue //hasn't moved recently
+		if(isrobot(M))
+			to_chat(world, "<span class='danger'>MOTION SENSOR DEBUG: Aborted: Subject is a robot.</span>")
+			continue
 		if(ishuman(M))
+			to_chat(world, "<span class='danger'>MOTION SENSOR DEBUG: Human detected.</span>")
 			var/mob/living/carbon/human/H = M
 			if(istype(H.wear_ear, /obj/item/device/radio/headset/almayer))
```
Supermichael777: Link this to ids so people wearing earmuffs don't set it off

PR #489, opened 2018-10-14
Makes bipods properly apply slow, made less clunky

Jalordon:
Fixes bipods so they apply slowdown the moment they are deployed, instead of applying the slowdown when the gun is wielded again.

Bipods are now allowed to be deployed without needing support and will no longer retract when firing without support. However, no benefit is applied unless firing while supported.


code/modules/projectiles/updated_projectiles/gun_attachables.dm
```diff
@@ -1037,16 +1037,16 @@ obj/item/attachable/attack_hand(var/mob/user as mob)
...
+					to_chat(user, "<span class='notice'>You deploy [src].</span>")
 					G.aim_slowdown += SLOWDOWN_ADS_SCOPE
 					G.wield_delay += WIELD_DELAY_FAST
-			else
-				to_chat(user, "<span class='notice'>There is nothing to support [src].</span>")
-				return FALSE
+				else
+					to_chat(user, "<span class='notice'>[src] is already deployed.</span>")
+	if (G.flags_item & WIELDED) //Updates slowdown, otherwise requires wielding again to update.
+		G.slowdown = G.aim_slowdown
```
Rohesie: It's `slowdown = initial(slowdown) + aim_slowdown` in the wielding proc. Weapons should start with zero initial slowdown, but in case that changes, it would be better. An update proc would be even better.

Rohesie:
You said you needed to make changes to this PR, in-game. Is there anything that is left to do?

Jalordon:
git incompetence, don't merge this until i get good and rebase properly

PR #492, opened 2018-10-15
Some new sprites for the Autodoc and the Bioprinter.

Ghommie:
Here to replace the old sprites. At least not in a codersprite way. 
Plus new icon_states for unpowered and busy.


Alterist:
Pretty good

PR #493, opened 2018-10-15
Revert "[100% tested] Ports multiple changes by goonstation to their process scheduler, These should cause a minor performance increase. (#339)

Rohesie:
Let's test if this was the cause of the terrible lag.


Ghommie:
Wouldn't it be more efficient to revert one at a time to find out the source of lag, especially from small pools?

PR #495, opened 2018-10-15
Adds in-game webmap button

AffectedArc07:
This PR adds an ingame UI button for the webmap, as suggested by Rohesie
image
Features:
- Will automatically select the correct ship and ground level
- Maps are config based, instead of being hardcoded in
- Button scales properly with correct anchoring

### DIFF
--- a/interface/interface.dm
+++ b/interface/interface.dm
@@ -53,6 +53,33 @@
 	src << link("https://github.com/ColonialMarines-Mirror/ColonialMarines-2018")
 	return
 
+/client/verb/webmap()
+	var/ship_link = ""
+	var/ground_link = ""
+	set name = "webmap"
+	set desc = "Opens the webmap"
+	set hidden = 1
+	var/choice = alert("Do you want to view the ground or the ship?",,"Ship","Ground","Cancel")
+	if(choice == "Cancel")
+		return
+	if(choice == "Ship")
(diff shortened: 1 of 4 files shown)


interface/interface.dm
```diff
@@ -53,6 +53,33 @@
...
+	set hidden = 1
+	var/choice = alert("Do you want to view the ground or the ship?",,"Ship","Ground","Cancel")
+	if(choice == "Cancel")
+		return
+	if(choice == "Ship")
+		if(MAIN_SHIP_NAME == "USS Almayer")
+			ship_link = config.almayer_url
+		src << link(ship_link)
+	if(choice == "Ground")
+		if(map_tag == "Ice Colony")
```
Rohesie: Can you do this under a `switch` ?

interface/interface.dm
```diff
@@ -53,6 +53,41 @@
...
 	return
 
+/client/verb/webmap()
+	var/ship_link
+	var/ground_link
+	set name = "webmap"
+	set desc = "Opens the webmap"
+	set hidden = 1
+	var/choice = alert("Do you want to view the ground or the ship?",,"Ship","Ground","Cancel")
+	if(choice == "Cancel")
```
Rohesie: Add a `switch` to these options and it's good.
You can run the "Cancel" simply as an `else` at the end.

PR #500, opened 2018-10-15
Xeno healing rework and fixes

Rohesie:
Adjust damage verbs were used only by simple mobs and xenos, though they follow the logic of humans in where double max health damage kills someone. Xenos have a different condition to die: less than critical health, which is `-100`. This caused the edge case of the weaker of xenos, the Young Runner, having 100 life and being able to take at most 200 damage of each kind. In order to die it would require to be under -100, that is why no matter how many bullets it took, it would stay alive, unless if it took 1 damage of another kind as well, like fire.
Larvas had a snowflake critical health, bypassing this.

* Added procs to adjust and reset pain vars.
* Added resetting of halloss, pain, stagger and slowdown vars, plus maximum plasma, to rejuvenate()
* Xenos no longer share the healing procs with simple mobs, to adjust for their specificity.
* Xeno healing proc tweaked a bit. It's easier to die if in crit and off-weeds, but recovery while on crit in weeds is much improved.
* Larvas heal fast anywhere, as long as not in crit and out of weeds.
* Fixes xenos being unable to heal without a living queen.
* Shooting at xeno's head no longer does +14% extra damage. That was a bad idea we inherited.
* Clean a bit of code
* Action button icons get updated in the life proc regardless of plasma amount.
* Only the drone tree can, again, pick facehuggers.
* Xenos without a queen can heal again.
* Pain vars and procs are for carbon, not for all the living.
* update_action_button_icons() defined for those that have actions

### DIFF
--- a/code/modules/mob/living/carbon/shock.dm
+++ b/code/modules/mob/living/carbon/shock.dm
@@ -1,12 +1,12 @@
 /mob/living/carbon/proc/getHalLoss()
 	return halloss
 
-/mob/living/carbon/proc/adjustHalLoss(var/amount)
+/mob/living/carbon/adjustHalLoss(amount)
 	if(status_flags & GODMODE)
 		return FALSE	//godmode
 	halloss = CLAMP(halloss+amount,0,maxHealth*2)
 
-/mob/living/carbon/proc/setHalLoss(var/amount)
+/mob/living/carbon/proc/setHalLoss(amount)
 	if(status_flags & GODMODE)
 		return FALSE	//godmode
 	halloss = amount
@@ -14,12 +14,12 @@
 /mob/living/carbon/proc/getTraumatic_Shock()
 	return traumatic_shock
 
-/mob/living/carbon/proc/adjustTraumatic_Shock(var/amount)
+/mob/living/carbon/proc/adjustTraumatic_Shock(amount)
 	if(status_flags & GODMODE)
 		return FALSE	//godmode
 	traumatic_shock = CLAMP(traumatic_shock+amount,0,maxHealth*2)
 
-/mob/living/carbon/proc/setTraumatic_Shock(var/amount)
+/mob/living/carbon/proc/setTraumatic_Shock(amount)
 	if(status_flags & GODMODE)
 		return FALSE	//godmode
 	traumatic_shock = amount
@@ -27,12 +27,12 @@
 /mob/living/carbon/proc/getShock_Stage()
 	return shock_stage
 
-/mob/living/carbon/proc/adjustShock_Stage(var/amount)
+/mob/living/carbon/proc/adjustShock_Stage(amount)
 	if(status_flags & GODMODE)
 		return FALSE	//godmode
 	shock_stage = CLAMP(shock_stage+amount,0,maxHealth*2)
 
-/mob/living/carbon/proc/setShock_Stage(var/amount)
+/mob/living/carbon/proc/setShock_Stage(amount)
 	if(status_flags & GODMODE)
 		return FALSE	//godmode
 	shock_stage = amount
--- a/code/modules/mob/living/carbon/shock.dm
+++ b/code/modules/mob/living/carbon/shock.dm
@@ -1,5 +1,5 @@
 /mob/living/var/traumatic_shock = 0
-/mob/living/carbon/var/shock_stage = 0
+/mob/living/var/shock_stage = 0
 
 // proc to find out in how much pain the mob is at the moment
 /mob/living/carbon/proc/updateshock()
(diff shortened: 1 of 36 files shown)


code/modules/mob/living/carbon/xenomorph/damage_procs.dm
```diff
@@ -108,13 +107,20 @@
 			adjustFireLoss(damage)
 
 	updatehealth()
-	return 1
+	return TRUE
+
+/mob/living/carbon/Xenomorph/adjustBruteLoss(var/amount)
```
ninjanomnom: You don't need `var/` in args like this

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
```diff
@@ -108,13 +107,20 @@
...
+	bruteloss = CLAMP(bruteloss + amount, 0, maxHealth - crit_health)
+
+/mob/living/carbon/Xenomorph/adjustFireLoss(var/amount)
```
ninjanomnom: ^

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -101,18 +95,40 @@ adjustFireLoss(-(maxHealth / 70 + 0.5 + (maxHealth / 70) * recovery_aura/2)*(m))
 		adjustFireLoss(fire_stacks + 3)
 
 /mob/living/carbon/Xenomorph/proc/handle_living_health_updates()
+	if(health >= maxHealth || hardcore) //no damage, don't bother
+		updatehealth()
```
ninjanomnom: Why update health if nothing is changing?

code/modules/mob/living/carbon/xenomorph/update_icons.dm
```diff
@@ -38,9 +35,13 @@
 			icon_state = "[caste] Running"
 		else
 			icon_state = "[caste] Walking"
-
 	update_fire() //the fire overlay depends on the xeno's stance, so we must update it.
 
+/mob/living/carbon/Xenomorph/proc/update_action_button_icons()
```
ninjanomnom: This seems like something worth having at the same place you can have actions instead of being a xenomorph specific proc

code/modules/mob/living/carbon/xenomorph/life.dm
Rohesie: `updatehealth()` does 3 things:
* Update health by recalculating it with the brute and fire loss.
* Update the medhud.
* Update the stat, killing the xeno if under a certain health, making it unconscious under another, and conscious over it.

It could be made more efficient, but the idea is that a xeno could transition from a state of low health to max, or one that warrants HUD changes. Suggestions?

code/modules/mob/living/carbon/xenomorph/update_icons.dm
Rohesie: Made an `item`-specific one, that was the only other instance using it.

code/modules/mob/living/carbon/shock.dm
```diff
@@ -1,5 +1,5 @@
 /mob/living/var/traumatic_shock = 0
-/mob/living/carbon/var/shock_stage = 0
+/mob/living/var/shock_stage = 0
```
Qustinnus: why are these vars in a carbon file if they're for living?

Qustinnus: and why are they for living? the procs are for carbon.

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
```diff
@@ -93,13 +94,11 @@
 	if(damage > 12) //Light damage won't splash.
 		check_blood_splash(damage, damagetype, chancemod)
 
-	if(stat == DEAD) return
+	if(stat == DEAD)
+		return FALSE
 
-	if(warding_aura && damage > 0) //Damage reduction. Every half point of warding decreases damage by 2.5 %. Maximum is 25 % at 5 pheromone strength.
```
Qustinnus: damage > 0 is useless

code/modules/mob/living/carbon/xenomorph/update_icons.dm
ninjanomnom: It doesn't matter what else is currently using it, it should be defined at whatever level can have action buttons

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
Rohesie: Yes, and it's removed.

PR #502, opened 2018-10-16
Armor Fixes

Alterist:
bits of code were commented out due to...fancy dividers?
fixes riot armors
makes marine armor not sometimes shitty for...reasons?

### DIFF
--- a/code/modules/clothing/suits/marine_armor.dm
+++ b/code/modules/clothing/suits/marine_armor.dm
@@ -17,7 +17,7 @@
 #endif
 
 //=======================================================================\\
-//=======================================================================\\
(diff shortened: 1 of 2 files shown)


code/modules/clothing/suits/marine_armor.dm
```diff
@@ -17,7 +17,7 @@
 #endif
 
 //=======================================================================\\
-//=======================================================================\\
+//=======================================================================
```
Rohesie: Interesting change

MoXuT0:
Doing the god's work. I always hated that there was 2 armor types without arms protection.

PR #504, opened 2018-10-16
Armor Reduction Fix - RNG Removal

Surrealaser:
-Armor now reduces damage by a % instead of having two ridiculous RNG rolls to determine whether or not damage is dealt or halved.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_damage.dm
+++ b/code/modules/mob/living/carbon/human/human_damage.dm
@@ -375,7 +375,7 @@ This function restores all limbs.
 	if(!organ)	return 0
 
 	if(blocked)
-		damage *= CLAMP(1-blocked,0.00,1.00) //Percentage reduction
+		damage *= CLAMP01(1-blocked,0.00,1.00) //Percentage reduction
 
 	switch(damagetype)
 		if(BRUTE)
--- a/code/modules/mob/living/carbon/human/human_damage.dm
+++ b/code/modules/mob/living/carbon/human/human_damage.dm
@@ -375,7 +375,7 @@ This function restores all limbs.
 	if(!organ)	return 0
 
 	if(blocked)
-		damage *= CLAMP(1-blocked,0.00,1.00) //Percentage reduction
+		damage *= CLAMP01(1-blocked) //Percentage reduction
 
 	switch(damagetype)
 		if(BRUTE)
--- a/code/modules/mob/living/carbon/human/human_damage.dm
+++ b/code/modules/mob/living/carbon/human/human_damage.dm
@@ -375,7 +375,7 @@ This function restores all limbs.
 	if(!organ)	return 0
 
 	if(blocked)
-		damage *= CLAMP01(1-blocked,0.00,1.00) //Percentage reduction
+		damage *= CLAMP(1-blocked,0.00,1.00) //Percentage reduction
 
 	switch(damagetype)
 		if(BRUTE)
--- a/code/modules/mob/living/carbon/human/human_damage.dm
+++ b/code/modules/mob/living/carbon/human/human_damage.dm
@@ -345,6 +345,12 @@ This function restores all limbs.
 
 	//visible_message("Hit debug. [damage]|[damagetype]|[def_zone]|[blocked]|[sharp]|[used_weapon]")
 
+	if(blocked >= 1) //Complete negation
+		return 0
+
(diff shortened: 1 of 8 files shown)


code/modules/mob/living/damage_procs.dm
```diff
@@ -9,21 +9,24 @@
 	standard 0 if fail
 */
 /mob/living/proc/apply_damage(var/damage = 0,var/damagetype = BRUTE, var/def_zone = null, var/blocked = 0, var/used_weapon = null, var/sharp = 0, var/edge = 0)
```
Ghommie: Also, wouldn't you mind cleaning up the code by removing all istances of the blocked var?

Ghommie:
Hold on for a moment here, man. There are a handful of armors with stats going over 100, could you please adjust those to avoid total immunity?

Surrealaser:
@Ghommie I can only assume those armours being that high is working as intended, and are intended to interact with armour piercing munitions and such. Both marines and aliens do have ways around this, and an armour rating of 100+ means you'd basically be immune to damage of the type it protects against in the old system.

code/modules/mob/living/living_defense.dm
```diff
@@ -11,25 +11,8 @@
...
-		return 2
-	if(absorb == 1)
-		if(absorb_text)
-			show_message("[soften_text]")
-		else
-			show_message("\red Your armor softens the blow!")
-		return 1
-	return 0
+	var/armor = CLAMP(1.00 - (getarmor(def_zone, attack_flag) / 100),0.00,1.00)
+	return armor
```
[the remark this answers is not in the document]
Surrealaser: Eh, I don't think it's necessary to pollute the logs with notifications that your armour did something when you now know it will always do something.

code/modules/mob/living/damage_procs.dm
Surrealaser: I'll look into it

Ghommie:
I don't find that concept appealing, but I do agree that armour piercing attacks can bypass it.

Surrealaser:
@Ghommie 100+ armor = damage invincibility whether in this system or the old. I don't like the existence of immunity stuff, but deciding whether or not they should exist is beyond the purview of this PR.

Surrealaser:
@Rohesie Well yes, but this is a good interim fix.

Rohesie:
It's not really a game-breaking issue that needs an urgent hotfix. I think we can think of a more palatable concept. Let's bring more people to this discussion.

Surrealaser:
@Rohesie People are complaining about the swingy, inconsistent damage constantly, not knowing that this shitty armour system is what's really going on.

code/modules/mob/living/carbon/human/human_damage.dm
```diff
@@ -345,6 +345,12 @@ This function restores all limbs.
 
 	//visible_message("Hit debug. [damage]|[damagetype]|[def_zone]|[blocked]|[sharp]|[used_weapon]")
 
+	if(blocked >= 1) //Complete negation
```
Rohesie:
```suggestion
	if(blocked || !damage) //Complete negation
```

Rohesie:
A fully RNG and a fully predictable system both have issues. A mix might be healthier. Let's take this for broader discussion.

code/modules/mob/living/carbon/human/human_damage.dm
```diff
@@ -371,7 +375,7 @@ This function restores all limbs.
 	if(!organ)	return 0
 
 	if(blocked)
-		damage = (damage/(blocked+1))
+		damage *= CLAMP(1-blocked,0.00,1.00) //Percentage reduction
```
Ghommie: `CLAMP01(1-blocked)`

code/modules/mob/living/carbon/human/human_damage.dm
Surrealaser: No, blocked has to be greater than 1, unless decimal amounts don't count as being true.

code/modules/mob/living/carbon/human/human_damage.dm
Rohesie:
```suggestion
		damage *= CLAMP01(1-blocked) //Percentage reduction
```

Surrealaser: This syntax actually works?

PR #507, opened 2018-10-16
New Yeezys (replacing CM boots with new sneakers)

Arachnidnexus:
New shoes for marines, replacing old CM sprites. Using color palette and base image of TG boots. yeezys

CM:
image
DMCA (heavily modified TG boots):
image


Qustinnus:
damn daniel

PR #509, opened 2018-10-17
Makes SMG AP ammo perform better

Jalordon:
-Damage changed from 20 - > 25 , for reference: regular smg ammo is 30 damage, regular M41 ammo is 40 damage and AP M41 ammo is 35 damage
-Scatter penalty on SMG AP ammo removed

Reasoning for change is based on local testing; SMG AP ammo was underperforming always compared to regular ammo, no matter if shooting low armor or high armored castes.


Jalordon:
obsolete if #510 merges

Rohesie:
#510 is merged.

PR #510, opened 2018-10-18
Attachment and Ammo Rebalance

Surrealaser:
SMG AP: 
-AP increased from 30 to 40.
-Scatter increment removed.
-This should make its performance better than standard vs enemies with 25+ armour, and worse vs enemies with 20 or less.

M44 Revolver:
-Damage reduced from 40 to 35.
-10 AP removed.
-Accuracy penalty removed.
-Fanning the hammer with the M44 should now still be quite good (and more consistent per the removal of accuracy penalties), but not to the point of doing excessive amounts of DPS/spike damage.

-M41 AP:
-Damage reduced from 35 to 30. This should make its performance better than standard vs enemies with 30+ armour, and worse vs enemies with less.

BFA and QFA:
-Accuracy penalty for each increased by +5 as they were too powerful when used in tandem with each other.

12 Gauge Slug:
-Stagger slow reduced from 4 to 2 until we have a better solution.


Rohesie:
As expected, this would conflict with the other.

PR #511, opened 2018-10-18
Defender: Fortify/Defensive Stance Improvements

Surrealaser:
-Fortify and Defensive Stance are now more of a toggle, and can be used/transitioned between while the other is active; you still cannot use other Defender abilities while either is active. There is a short 2 second cooldown before you can toggle on/off either.

-You can now remain in Fortify stance indefinitely.

-Fortify grants Crusher level explosive resistance and grants immunity to staggerstun (such as from slugs) while it's active.

-Defensive stance halves the effect of staggerstun and negates its stun.

### DIFF
--- a/code/modules/projectiles/updated_projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/updated_projectiles/ammo_datums.dm
@@ -110,12 +110,23 @@
 		if(!isliving(M))
 			return
 		var/impact_message = ""
+		if(isXeno(M))
+			var/mob/living/carbon/Xenomorph/D = M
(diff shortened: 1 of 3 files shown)


code/modules/projectiles/updated_projectiles/ammo_datums.dm
```diff
@@ -110,12 +110,23 @@
 		if(!isliving(M))
 			return
 		var/impact_message = ""
+		if(isXeno(M))
```
Rohesie: This would be ideally handled by procs specific to each mob, something like `on_hit`, instead of a long list of checks and exceptions.

PR #512, opened 2018-10-18
Melee Tweak

Surrealaser:
-Machete attack damage increased by +5 to 40, and attack delay decreased to 9 down from 11 (about a 22% increase in attack speed). DPS increased from ~32 to ~49.

-Combat knife attack delay reduced from 11 to 7 (~57.14% increase in attack speed) . DPS increased from ~23 to 35.71

-Effect/impact of the changes is to make pure melee a little more competitive vs bayos, and to improve the knife and machete's niches as weed clearers.


Rohesie:
Good concept. Let's see how it fares in tests.

PR #513, opened 2018-10-18
Random Sprite Stuff

Alterist:
ported some sprites that might not be currently used, but simply have an better sprite on another open codebase
moved some files together
deleted some WY references


Rohesie:
To be fair I like the old welder and weldertank, as well as the metal sink.
Regarding the Christmas fluff, are these in the to-be-removed list?

Alterist:
metal sink is changed because sink and sink_alt had the same sprite
Christmas stuff was deleted because there's a copy of the same file in icons/seasonal

PR #520, opened 2018-10-20
Pump Notification Spam Fix

Surrealaser:
Per title


Supermichael777:
adds spam

code/modules/projectiles/updated_projectiles/guns/shotguns.dm
```diff
@@ -428,12 +428,12 @@ can cause issues with ammo types getting mixed up during the burst.
 
 //More or less chambers the round instead of load_into_chamber(). Also ejects used casings.
 /obj/item/weapon/gun/shotgun/pump/proc/pump_shotgun(mob/user)	//We can't fire bursts with pumps.
+	if(world.time < (recent_pump + pump_delay) ) //Don't spam it.
+		return
 	if(pump_lock)
 		playsound(user,'sound/weapons/throwtap.ogg', 25, 1)
 		to_chat(user,"<span class='warning'><b>[src] has already been pumped, locking the pump mechanism; fire or unload a shell to unlock it.</b></span>")
 		return
```
Rohesie:
```suggestion
		recent_pump = world.time
		return
```

Surrealaser: That's probably not the best idea, as you're refreshing the delay without actually pumping the gun.

Rohesie: It's just a small delay, and if you don't you get spammed by those warnings.

Supermichael777: I actually agree with surrealaser, that's counter-intuitive and will fuck over people in combat mashing to make their foes horizontal.

Rohesie: It won't affect them. It will only refresh the delay if you have already pumped. It won't make people pump any slower.

Surrealaser: Added a notification cooldown.

PR #521, opened 2018-10-20
Smartgunner Arm-or

InterroLouis:
Give smartgunner arm armor. On their arms.


Alterist:
So what you're saying is, Smartgunner armor didn't have arms, and you are giving them arm armor on their arms, thus buffing the armor?

InterroLouis:
Yes, the idea is to give them armor on their arms. I'm gonna go ahead and reference your own PR, #502 where you made the M3 Pattern armors that visually don't have arms, give arm protection.

Supermichael777:
because its so easy to disable their two handed weapon with a fucking young runner slash to the arm

factoryman942:
SG armour is mandatory to use the gun and it gives the SG user a specific weakness (remove arm, no more dakka). The PFC armour was completely random visual variations.

PR #530, opened 2018-10-21
Ladder bug fix.

Rohesie:
- Removed the limitation of one user per ladder. Supposed realism (questionable) here detracts from balance (worsening ladder camping/cheesing) and, more importantly, from fun (having to wait until someone else finishes using a ladder).
- Made a proc to handle teleporting mobs while preserving grabs, teleporting all the unanchored objects and mobs associated, unbuckling them if necessary.

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -647,4 +647,82 @@ mob/proc/yank_out_object()
 	if(istype(AM, /obj/item))
 		temp_drop_inv_item(AM, TRUE) //unequip before deletion to clear possible item references on the mob.
 
+/mob/forceMove(atom/destination)
+	stop_pulling()
+	if(pulledby)
+		pulledby.stop_pulling()
+	if(buckled)
+		buckled.unbuckle()
+	. = ..()
+	if(.)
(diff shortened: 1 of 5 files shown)


ninjanomnom:
This code change shouldn't have any effect, are you sure this is fixing anything?

ninjanomnom:
Yeah a runtime error there sounds like the more likely answer. The code changes you've done here should have no difference in functionality

Rohesie:
Alright, that should be handled in this solution, as well.

code/modules/mob/mob.dm
```diff
@@ -647,4 +647,82 @@ mob/proc/yank_out_object()
...
+			break
+		conga_line += A
+		var/mob/M = A
+		if(istype(M)) //Is a mob
+			var/mob/living/L = M
+			if(istype(L))
+				L.smokecloak_off()
+			if(M.buckled && !(M.buckled in conga_line))
+				if(M.buckled.anchored)
+					M.buckled.unbuckle() //Force the dragged people out of anchored buckles.
```
Qustinnus: Just end it at buckled people on anchored objects

PR #539, opened 2018-10-22
Admin: Proper table slam, window slam, swirlie, toilet slam logging

DominikPanic:


Rohesie:
I'm not sure if the admins need to be messaged each time this happens, regardless of who does it.

subject217:
yeah, it should be in attack logs

DominikPanic:
Msg_admin_attack is called for every grab, punch, etc so I doubt spamming would be an issue as that log is already meant to be spammy instead of the FF one.

PR #553, opened 2018-10-23
Tank Bumping

JeserTheFox:
1. All xenos and humans get thrown at random direction when tank bumps into them.
2. Everyone get very short knockdown and small damage, except for Crusher and Queen, they get neither.
[Visual demonstration.](https://i.imgur.com/7kWPDOJ.gifv)


code/modules/vehicles/cm_armored.dm
```diff
@@ -368,15 +368,28 @@ var/list/TANK_HARDPOINT_OFFSETS = list(
...
+			temp = get_step(T, facing)
+			T = temp
+			T = get_step(T, pick(cardinal))
+			if(M.mob_size == MOB_SIZE_BIG)
+				M.throw_at(T, 2, 1, src, 0)
+			else
+				M.throw_at(T, 2, 1, src, 1)
+			M.KnockDown(1)
+			M.apply_damage(10 + rand(0, 5), BRUTE)
+			return
```
Rohesie: This `return` prevents the `visible_message` from appearing.
From what the logic of this tells, it will only appear for lying mobs and larvas, who wouldn't actually be bumped.

JeserTheFox: By current logic, no one will see message at all. Gonna fix that.

PR #556, opened 2018-10-23
CIC Rework and some other stuff

Looreydev:
- Reworked the CIC a little bit
- Changed XOs headwear to a Beret and changed something in supply exports


code/datums/supplyexports.dm
```diff
@@ -10,7 +10,7 @@
 
 
 /datum/supply_export/xemomorph/larva
-	cost = 0
+	cost = 3
```
Rohesie: Undocumented change. Also undesirable for larvas to cost more than a single point, really. We'd have to check for embryos as well.

Rohesie:
Atomize your PRs whenever you can.

Looreydev:
Somehow broke the map, gonna get myself the normal version of the Almayer and try editing it again.

PR #456, opened 2018-10-11
Hypospray Improvement

Surrealaser:
-Hypospray now has a translucent segment where you can see how much chem is loaded into it.

-Examining the hypospray now shows you its dosage level, as well as its contents by reagent, including their U, % of the total, and amount of each chem transferred per dose.

-Using the hypospray now sets its label to whatever you like. It can only have one label in this way at a time.

-Hypospray injection amounts/transfer rates are now customizable.

-Hypospray capacity increased to 60U.

-The hypospray medics start with in their essential kit now comes pre-loaded with 60u Oxycodone; use it wisely.

PR #557, opened 2018-10-23
FIXES NICE COLONY

Alterist:
fuck


Alterist:
Makes Ice load with  #456

PR #560, opened 2018-10-24
Decapitate Decapitated

Surrealaser:
-No more RNG chance of decapitation from damage if wearing a helmet, and not a synth.

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -225,7 +225,9 @@
 	src.update_damages()
 
 	//If limb took enough damage, try to cut or tear it off
-	if(body_part != UPPER_TORSO && body_part != LOWER_TORSO && !no_limb_loss)
+	var/obj/item/clothing/worn_helmet = owner.head
+	if(body_part != UPPER_TORSO && body_part != LOWER_TORSO && !no_limb_loss && (body_part != HEAD && !istype(worn_helmet, /obj/item/clothing/head/helmet) && !owner.species.flags & IS_SYNTHETIC)) //Synth heads can come off, not human ones for gameplay/balance reasons... if they're wearing a helmet.
+
 		if(config.limbs_can_break && brute_dam >= max_damage * config.organ_health_multiplier)
 			var/cut_prob = brute/max_damage * 10
 			if(prob(cut_prob))
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -227,9 +227,11 @@
 	//If limb took enough damage, try to cut or tear it off
 
 	if(body_part == UPPER_TORSO || body_part == LOWER_TORSO || no_limb_loss)
+		owner.updatehealth()
 		return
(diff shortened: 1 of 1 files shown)


code/modules/organs/limbs.dm
```diff
@@ -225,7 +225,9 @@
 	src.update_damages()
 
 	//If limb took enough damage, try to cut or tear it off
-	if(body_part != UPPER_TORSO && body_part != LOWER_TORSO && !no_limb_loss)
+	var/obj/item/clothing/worn_helmet = owner.head
+	if(body_part != UPPER_TORSO && body_part != LOWER_TORSO && !no_limb_loss && (body_part != HEAD && !istype(worn_helmet, /obj/item/clothing/head/helmet) && !(owner.species.flags & IS_SYNTHETIC))) //Synth heads can come off, not human ones for gameplay/balance reasons... if they're wearing a helmet.
```
Rohesie: Current logic doesn't work, try an early return and a more clear check.

Ghommie:
I'd say it's preferable to check for the headwear melee protection rather than the path's, and the decap chance lowering as the stat raises, with the 0% point at about the standard marine's helmet stat.

Surrealaser:
> It's preferable to check for the headwear melee protection rather than the path's, so to include all sufficiently armored headwear, not just types.
> Optionally, have the decap chance lowering as the stat raises, with the 0% point at about the standard marine's helmet stat.

To be clear, I am prioritizing balance and fun over realism here, hence using only the helmet path.

Rohesie:
Ghommie's suggestion preserves balance, with helmets providing 100% defense against decaps.

Surrealaser:
> Ghommie's suggestion preserves balance, with helmets providing 100% defense against decaps.

Put it this way: I'm not sure why the added complexity is needed for an ultra niche category of helmets that basically never see use.

code/modules/organs/limbs.dm
```diff
@@ -225,12 +225,17 @@
...
 	//If limb took enough damage, try to cut or tear it off
-	if(body_part != UPPER_TORSO && body_part != LOWER_TORSO && !no_limb_loss)
-		if(config.limbs_can_break && brute_dam >= max_damage * config.organ_health_multiplier)
-			var/cut_prob = brute/max_damage * 10
-			if(prob(cut_prob))
-				droplimb()
-				return
+
+	if(body_part == UPPER_TORSO || body_part == LOWER_TORSO || no_limb_loss)
+		return
```
Rohesie: Update health before you return.

code/modules/organs/limbs.dm
```diff
@@ -225,12 +225,17 @@
...
+	var/obj/item/clothing/worn_helmet = owner.head
+	if(body_part == HEAD && istype(worn_helmet, /obj/item/clothing/head/helmet) && !(owner.species.flags & IS_SYNTHETIC) ) //Early return if the body part is a head but target is wearing a helmet and is not a synth
+		return
```
Rohesie: Here as well.

PR #536, opened 2018-10-22
Fixes a runtime error and cleans the code a little in medical_tools.dm

Rohesie:
`reagents.set_reacting()` was being called before `create_reagents()`, thus accessing a proc inside a null variable/object. Blame #417

PR #567, opened 2018-10-24
Faulty code fixing.

Ghommie:
More bug fixes related to my doing.
Includes:
-Runtimes related to hud updates on clientless mobs.
-A runtime on the old mech syringe gun. EDIT: reverted in favor of #536 
-dead mobs possibly not processing update_cloak() and remaining invisible.

Also translocated those set invisibility/set background procs shared among carbon lifeforms' to the parent's life().


code/modules/mob/living/living.dm
```diff
@@ -1,6 +1,8 @@
 /mob/living/Life()
 	. = ..()
 
+	update_cloak()
```
Rohesie: Put this in `update_stat` or something like that, please.

PR #569, opened 2018-10-25
Overwatch and Laser Target reworks

Rohesie:
- Overwatch divided in four squad consoles and a main one.
- Squad consoles change leaders, set objectives, handle supply drops and the rail gun.
- Main console can see and message all squads and handles the orbital canon.
- Rail gun added, a little peashooter for SOs to pew-pew through lazes and orbital beacons. Weak, but at least it's one shot per minute.
- Tactical binocular's laze can be observed by overwatches and used as a target for the orbital canon.
- Orbital Beacons repurposed. Xenos can put acid to destroy them, and they can be planted and removed. They have an integrated camera to allow overwatches to see through them and work as targets for the orbital canon, not being destroyed in the process.
- Fixes a bug in where people who went to cryo appeared as dead.
- Gibbed marines are now reported as FUBAR, not merely DEAD.
- Tactical Lasers get slightly more informative names and an actual unique ID.


code/modules/cm_marines/overwatch.dm
```diff
@@ -1,59 +1,92 @@
+#define OW_MAIN 0
+#define OW_MONITOR 1
+#define OW_SUPPLIES 2
+#define OW_BOMBS 3
```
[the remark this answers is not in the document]
Rohesie: Done.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -52,15 +52,25 @@
...
-		assigned_squad = null
+	clean_from_squad(FALSE)
 	remove_from_all_mob_huds()
-	. = ..()
+	return ..()
+
+/mob/living/carbon/human/proc/clean_from_squad(wipe = FALSE)
+	if(!assigned_squad)
+		return FALSE
+	var/n = assigned_squad.marines_list.Find(src)
```
ninjanomnom: Make marines_list an associative array in the format human:TRUE and do `assigned_squad.marines_list -= src` instead

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -52,15 +52,25 @@
...
-			assigned_squad.marines_list[n] = name //mob reference replaced by name string
-		if(assigned_squad.squad_leader == src)
-			assigned_squad.squad_leader = null
-		assigned_squad = null
+	clean_from_squad(FALSE)
 	remove_from_all_mob_huds()
-	. = ..()
+	return ..()
+
+/mob/living/carbon/human/proc/clean_from_squad(wipe = FALSE)
```
ninjanomnom: This would make a lot more sense as a proc on the squad object

PR #571, opened 2018-10-25
Adding my WIP Map Hyperia Laboratories with all the files it needs

io-util:
- Adding the Map file For Hyperia Laboratories to the WIP folder (feel free to check it out **ITS WIP**)

- Adding a Area file for the map that goes by the name "HyperiaLabs.dm"

- Editing the area.dmi file to get the appropriate icons for my areas inside of it

- Editing another area file to add the 2 Drop Ship zones

(Wanted to make a PR so everyone is freely able to look at it, suggest changes and _maybe_ also edit it which i dont suggest because im actively working on it)


Looreydev:
It does compile just fine and i also tested it myself its playable (kind of) but of course its not the final map as it should be, just wanted to add it.

Ghommie:
Umh, just from a first very basic and quick look, only the LZ/hangar areas have a metal ceiling, and that generally makes mortars, CAS, OBs and medevac stretchers less practicable. But we'll see.

Looreydev:
Well yeah I only made it so the Hangar areas have Glass Ceilings RN and everything else is metal, as I said only feedback from other peeps can tell what needs to be changed and where

Supermichael777:
Okay so first this map is physically quite small, almost skeletal.

It also stuffed with long one tile paths (Horrible for both sides, as they turn into meat-grinders and sniper alleys.) and single entrance rooms (Horrible deathtraps for marines as they have no flank to push).

Almost all the walls lead into space so you cant flank through them.  That's not a bad way to make an area defensible but its simply EVERYWHERE.

PR #572, opened 2018-10-25
Facehugger/Carrier Tweak

Surrealaser:
-Facehugger knockout time reduced to 10 seconds down from 30.

-Facehuggers no longer knock off helmets for EZ decaps (they still damage hugger resistant helms such that they can no longer block huggers).  If you're afraid of camera helms, huggers knock them out when they successfully face fuck someone. You can also use your ability to knock out lights so that their non-NV lenses can't see shit. 

-Compensation: Carrier facehugger throw speed doubled. Facehugger throw cooldown decreased to 25 > 20 > 15 > 10 from 30 > 30 > 20 > 10.

-Reduced Facehugger latency on direct hits with thrown Facehuggers.

-Carriers can now vent crawl; this should help them be more offensive and help with weeding.

-Facehugging someone now takes half a second; this should significantly curtail combat hugging.


code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
```diff
@@ -92,6 +92,11 @@
 	return FALSE // Else you can't pick.
 
 /obj/item/clothing/mask/facehugger/attack(mob/M, mob/user)
+	user.visible_message("<span class='warning'>\ [user] attempts to plant [src] on [M]'s face!</span>", \
+	"<span class='warning'>You attempt to plant [src] on [M]'s face!</span>")
+	if(ishuman(M)) //Delay for humans only, not monkeys and shit
```
Rohesie:
```suggestion
	if(M.client && M.stat) //Delay for conscious cliented mobs, who should be resisting.
```

code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
```diff
@@ -92,6 +92,11 @@
...
+		if(!do_after(user, 5, TRUE, 5, BUSY_ICON_HOSTILE))
+			return
 	if(CanHug(M))
```
Rohesie: This could go before, as an early return. Better than adding a delay before informing them it can't be attached.

code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
Surrealaser: This broke the delay somehow

PR #576, opened 2018-10-26
Sentry Upgrades + Minisentries - READY FOR TEST MERGE

Surrealaser:
-Sentry projectiles made slightly more powerful and accurate; 40 damage up from 35; penetration remains at 20.

-Sentry burst now customizable from 2 to 6 shots. Accuracy loss and scatter scale with the number of shots. Delay between bursts shots reduced to 0.1.

-Sentry no longer tries to acquire targets it has no clear line of fire to.

-Sentry now has a 360 mode that reduces its detection radius from 7 to 3 tiles.

-Sentry now has an alert notification system that triggers while on when it detects a hostile, falls over, takes damage, is destroyed, or runs out of ammo.

-Dropship gauss turret damage and pen increased considerably. These are still pretty easy to juke out and flank, but now should be feared.

-A big limitation is this thing's reliance on a slow process, making it incredibly sluggish; when the detpacks are merged, I will probably modify this to tap into its faster timer process.

-Still waiting on proper hand/back sprites though.

-Minisentries have 150 HP vs 200 HP of the standard sentry, have a lower knockover threshold of 70 damage vs 100, and generally have their maintenance actions (battery/ammo replacement) done twice as quickly.

-Minisentries can be point bought from Engi vendors for 26 points, or be bought at Req for 80 requisition.

-Minisentries have no manual fire mode; they are auto-fire only, and have a smaller burst ceiling than the standard sentry; 2-4 vs 2-6.

-Minisentries can be folded up, or deployed from their folded state; this takes about 4 seconds. A folded Minisentry can fit on the back, or in a technician's backpack (which can now hold M56D and sentry parts).


code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -292,6 +297,11 @@
 	var/immobile = 0 //Used for prebuilt ones.
 	var/datum/ammo/bullet/turret/ammo = /datum/ammo/bullet/turret
 	var/obj/item/projectile/in_chamber = null
+	var/alerts_on = TRUE
+	var/last_alert = 0
+	var/alert_delay = 20 //20 seconds
```
Rohesie:
```suggestion
	#define TURRET_ALERT_DELAY 20 //20 seconds
```

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -477,6 +490,36 @@
...
 
+		if("toggle_alert")
+			if(!alerts_on)
+				user.visible_message("<span class='notice'>[user] activates [src]'s alert notifications.</span>",
+				"<span class='notice'>You activate [src]'s alert notifications.</span>")
+				visible_message("\icon[src] <span class='notice'>The [name] buzzes in a monotone voice: 'Alert notification system initiated'.</span>'")
+				alerts_on = TRUE
+				update_icon()
+			else
+				alerts_on = FALSE
```
Rohesie: This could be condensed with `alerts_on = !alerts_on` and `alerts_on ? this : that`

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -477,6 +490,36 @@
...
+				user.visible_message("<span class='notice'>[user] deactivates [src]'s alert notifications.</span>",
+				"<span class='notice'>You deactivate [src]'s alert notifications.</span>")
+				visible_message("\icon[src] <span class='notice'>The [name] buzzes in a monotone voice: 'Alert notification system deactivated'.</span>'")
+				update_icon()
+
+		if("toggle_radial")
+			if(!alerts_on)
```
Rohesie: I assume here you wanted `if(!radial_mode)`, but you can also apply the same mentioned logic.

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -801,6 +846,10 @@
 		update_icon()
 		return
 
+	//Clear the list after the delay
+	if(world.time > last_alert + alert_delay)
```
Rohesie:
```suggestion
	if(world.time > last_alert + TURRET_ALERT_DELAY)
```

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -931,6 +1001,16 @@
 		var/mob/living/carbon/human/H = M
 		if(istype(H) && H.get_target_lock(iff_signal)) continue
 
+
+
+		if(alerts_on)
+			var/list_match = alert_list[M]
+			if((world.time > last_alert + alert_delay) || !list_match) //if we're not on cooldown or the target isn't in the list, sound the alarm
```
Rohesie:
```suggestion
			if((world.time > last_alert + TURRET_ALERT_DELAY) || !list_match) //if we're not on cooldown or the target isn't in the list, sound the alarm
```

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -939,7 +1019,13 @@
 
 		if(path.len)
 			for(T in path)
-				if(T.opacity) continue
+				if(T.opacity || T.density) continue //don't pass on targets we can't actually fire at.
```
Rohesie:
```suggestion
				if(T.opacity || T.density)
					continue //don't pass on targets we can't actually fire at.
```

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -939,7 +1019,13 @@
...
+			var/obj/machinery/MA
+			for(MA in path)
+				if(MA.opacity || MA.density) continue //don't pass on targets we can't actually fire at.
```
Rohesie:
```suggestion
				if(MA.opacity || MA.density)
				 continue //don't pass on targets we can't actually fire at.
```

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -939,7 +1019,13 @@
...
+			var/obj/structure/S
+			for(S in path)
+				if(S.opacity || S.density) continue //don't pass on targets we can't actually fire at.
```
Rohesie:
```suggestion
				if(S.opacity || S.density)
				 continue //don't pass on targets we can't actually fire at.
```

code/modules/cm_marines/equipment/sentries.dm
Rohesie: What about barricades?

code/modules/projectiles/updated_projectiles/ammo_datums.dm
```diff
@@ -792,15 +792,20 @@
...
-	accuracy = -config.min_hit_accuracy
 
 /datum/ammo/bullet/turret/dumb
 	icon_state 	= "bullet"
 	iff_signal = 0
 
+/datum/ammo/bullet/turret/gauss
+
+/datum/ammo/bullet/turret/gauss/New()
+	..()
```
Rohesie:
```suggestion
	. = ..()
```

code/modules/cm_marines/equipment/sentries.dm
Qustinnus: alert delay can be a variable because you might want to change it on other types. Also the define shouldnt go here.

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -1,6 +1,10 @@
 //Deployable turrets. They can be either automated, manually fired, or installed with a pAI.
 //They are built in stages, and only engineers have access to them.
 
+#define SENTRY_ALERT_AMMO			0
+#define SENTRY_ALERT_HOSTILE		1
+#define SENTRY_ALERT_FALLEN			2
```
Qustinnus: Put these in _Defines. Using defines in a place like this might mean theyre defined after code that uses them wants to use them. (And these seem like defines you might want to know outside of this file)

code/modules/cm_marines/equipment/sentries.dm
Rohesie: The define shouldn't go here for sure, but so far this is a constant, not a variable. It's not changed anywhere else.

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -931,16 +1016,45 @@
...
+			path = getline2(src, M, TRUE)
 		else
 			continue
 
 		if(path.len)
+			var/blocked = FALSE //reset blocked
 			for(T in path)
-				if(T.opacity) continue
-			targets += M
+				if(T.opacity || T.density)
```
Rohesie: Won't this prevent firing over barricades?

Surrealaser: Fixed

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -1092,10 +1262,197 @@
...
+	if(!ishuman(usr))
+		return
+	var/turf/target = get_step(user.loc,user.dir)
+	if(!target)
+		return
+	var/blocked
+	if(target.density)
+		blocked = TRUE
+	if(!blocked) //allows us to stop checks and thus save resources if we find something that blocks placement
+		if(locate(/obj/machinery) in target)
```
Ghommie: var/blocked = TRUE
	if(!target.density)
		for(var/obj/O in target)
			if(O.density && !(S.flags_atom & ON_BORDER))
				blocked = TRUE
				break
	else
		blocked = TRUE

Ghommie: whoops, I didn't mean to default the var to true, fixed it for ya

Surrealaser: I was deliberately cutting down on categories of objects to the two that can actually block LoF to save resources

maps/Z.03.USS_Almayer.dmm
```diff
@@ -4144,7 +4144,7 @@
...
 "bBN" = (/obj/structure/table/almayer,/obj/item/storage/firstaid/rad,/obj/machinery/light{icon_state = "tube1"; dir = 8},/turf/open/floor/almayer{icon_state = "plate"; tag = "icon-sterile"},/area/almayer/engineering/lower_engine_monitoring)
-"bBO" = (/obj/structure/table/almayer,/obj/item/storage/pill_bottle/antitox{pixel_x = -5},/obj/item/storage/pill_bottle/antitox,/obj/item/storage/pill_bottle/bicaridine{pixel_x = 5},/obj/machinery/firealarm{pixel_y = 28},/turf/open/floor/almayer{tag = "icon-sterile_green_side (NORTH)"; icon_state = "sterile_green_side"; dir = 1},/area/almayer/medical/lower_medical)
+"bBO" = (/obj/structure/table/almayer,/obj/item/storage/pill_bottle/dylovene{pixel_x = -5},/obj/item/storage/pill_bottle/dylovene,/obj/item/storage/pill_bottle/bicaridine{pixel_x = 5},/obj/machinery/firealarm{pixel_y = 28},/turf/open/floor/almayer{tag = "icon-sterile_green_side (NORTH)"; icon_state = "sterile_green_side"; dir = 1},/area/almayer/medical/lower_medical)
```
Rohesie: Merge trash?

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -931,16 +1063,47 @@
 		var/mob/living/carbon/human/H = M
 		if(istype(H) && H.get_target_lock(iff_signal)) continue
```
Rohesie:
```suggestion
		if(istype(H) && H.get_target_lock(iff_signal))
		 continue
```

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -931,16 +1063,47 @@
...
 			for(T in path)
-				if(T.opacity) continue
-			targets += M
+				if(T == current_turf) //Don't deal with the current turf
+					continue
+				if(T.opacity || T.density)
+					blocked = TRUE
+					continue //don't pass on targets we can't actually fire at.
+
+				if(!blocked) //allows us to stop checks and thus save resources if we find something that blocks LoS/LoF
```
Rohesie: There is a `continue`, after `blocked = TRUE`, so this check will never fail and is thus not necessary.

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -931,16 +1063,47 @@
...
+					if(locate(/obj/machinery) in T)
+						var/obj/machinery/MA
+						for(MA in T)
+							if(MA.opacity || MA.density && !(MA.flags_atom & ON_BORDER))
+								blocked = TRUE
+								continue //don't pass on targets we can't actually fire at.
+
+				if(!blocked) //allows us to stop checks and thus save resources if we find something that blocks LoS/LoF
```
Rohesie: Same here.

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -931,16 +1063,47 @@
...
+					if(locate(/obj/structure) in T)
+						var/obj/structure/S
+						for(S in T)
+							if(S.opacity || S.density && !(S.flags_atom & ON_BORDER) )
+								blocked = TRUE
+								continue //don't pass on targets we can't actually fire at.
+			if(!blocked)
```
Rohesie: And here.

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -1092,10 +1254,201 @@
...
+	health_max = 155
+	rounds = 500
+	rounds_max = 500
+	knockdown_threshold = 70 //lighter, not as well secured.
+	work_time = 20 //significantly faster than the big sentry
+	ammo = /datum/ammo/bullet/turret/mini //Similar to M39 AP rounds.
+	magazine_type = /obj/item/ammo_magazine/minisentry
+
+/obj/item/storage/box/sentry/New()
+	..()
```
Rohesie:
```suggestion
	. = ..()
```

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -1092,10 +1254,201 @@
...
+	update_icon()
+
+/obj/machinery/marine_turret/mini/MouseDrop(over_object, src_location, over_location) //Drag the tripod onto you to fold it.
+	if(!ishuman(usr))
+		return
+	var/mob/living/carbon/human/user = usr //this is us
+	if(over_object == user && in_range(src, user))
```
Rohesie: Early return.

code/modules/cm_marines/equipment/sentries.dm
Rohesie: I'd have to check if `locate(type) in list` saves resources before a `for(type in list)`

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -1092,10 +1254,201 @@
...
+	icon_state = "sentry_case"
+	w_class = 5
+	storage_slots = 4
+	can_hold = list(/obj/item/device/marine_turret/mini, //gun itself
+					/obj/item/tool/wrench, //wrench to hold it down into the ground
+					/obj/item/tool/screwdriver, //screw the gun onto the post.
+					/obj/item/ammo_magazine/minisentry)
+
+/obj/item/storage/box/minisentry/New()
+	..()
```
Rohesie:
```suggestion
	. = ..()
```

code/modules/projectiles/updated_projectiles/ammo_datums.dm
```diff
@@ -792,15 +792,29 @@
...
 	iff_signal = 0
 
+/datum/ammo/bullet/turret/gauss
+
+/datum/ammo/bullet/turret/gauss/New()
+	. = ..()
+	damage = config.med_hit_damage
+	penetration= config.mhigh_armor_penetration
+
+/datum/ammo/bullet/turret/mini
```
Rohesie:
```suggestion
/datum/ammo/bullet/turret/mini
	name = "turret slimline bullet"
```
Or something like that.

code/modules/projectiles/updated_projectiles/ammo_datums.dm
```diff
@@ -792,15 +792,29 @@
...
+	damage = config.lmed_hit_damage
 	penetration= config.low_armor_penetration
-	accuracy = -config.min_hit_accuracy
+	damage_falloff *= 0.5 //forgot to add this
 
 /datum/ammo/bullet/turret/dumb
 	icon_state 	= "bullet"
 	iff_signal = 0
 
+/datum/ammo/bullet/turret/gauss
```
Rohesie:
```suggestion
/datum/ammo/bullet/turret/gauss
	name = "turret heavy bullet"
```
Or something like that.

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -700,34 +814,30 @@
...
+				sentry_alert(SENTRY_ALERT_FALLEN)
 			on = FALSE
-	if(stat)
-		density = FALSE
-	else
-		density = initial(density)
 	update_icon()
 
 /obj/machinery/marine_turret/proc/check_power(var/power)
 	if (!cell)
```
Rohesie:
```suggestion
	if (!cell || !on || stat)
```

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -700,34 +814,30 @@
...
-		icon_state = "sentry_battery_none"
+		update_icon()
 		return FALSE
 
 	if(!on || stat)
```
Rohesie:
```suggestion
```

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -700,34 +814,30 @@
...
-		on = FALSE
-		icon_state = "sentry_off"
+		update_icon()
```
Rohesie:
```suggestion
```

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -700,34 +814,30 @@
...
 		return FALSE
```
Rohesie:
```suggestion
```

code/modules/cm_marines/equipment/sentries.dm
```diff
@@ -918,29 +1050,57 @@
...
 		else
 			continue
 
 		if(path.len)
+			var/blocked = FALSE
+			var/current_turf = get_turf(src)
 			for(T in path)
-				if(T.opacity) continue
-			targets += M
+				if(T == current_turf) //Don't deal with the current turf
```
Rohesie: `path -= current_turf` before the `for` loop should offer better performance. Does the check only once.

Rohesie: Or better even, after `path = getline2(src, M, TRUE)`

code/modules/cm_marines/equipment/sentries.dm
Surrealaser: This has been done.

code/modules/cm_marines/equipment/sentries.dm
Surrealaser: This has been done.

code/modules/cm_marines/equipment/sentries.dm
Surrealaser: I added this because an earlier change I made to getline2 didn't seem to work, but upon further testing it apparently does; this is handled.

PR #579, opened 2018-10-27
Adds 3 New UI Presets

Suicidalpickles:
ports in the clockwork, operative, and slimecore UI's from /tg/
all can be selected by the typical means of character prefs

Operative:
capture
Clockwork:
capture2
Slimecore:
capture3


Supermichael777:
Slime-core has always been a good balance between at a glance slot recognition and not blinding me.

PR #580, opened 2018-10-27
Admin: Better autoresponses, Proper re-adminning, Split ahelps into adminhelps and mentorhelps, suicide logging fix

DominikPanic:


code/modules/admin/holder2.dm
```diff
@@ -26,14 +26,18 @@ var/list/admin_datums = list()
 	admin_datums[ckey] = src
 
 /datum/admins/proc/associate(client/C)
+	to_chat(usr, "associate triggered")
 	if(istype(C))
+		to_chat(usr, "client works")
```
Rohesie: Are these intended to remain or are these debug messages only?

code/modules/admin/verbs/adminhelp.dm
```diff
@@ -107,42 +113,50 @@ var/list/adminhelp_ignored_words = list("unknown","the","a","an","of","monkey","
...
 	msg = "<br><br><font color='#009900'><b>[selected_upper]: [get_options_bar(mob, 2, 1, 1)]:</b></font> <br><font color='#DA6200'><b>[msg]</font></b><br>"
 
 	var/admin_number_afk = 0
 
 	var/list/mentorholders = list()
-	var/list/debugholders = list()
 	var/list/adminholders = list()
 	for(var/client/X in admins)
-		if(R_MENTOR & X.holder.rights && !(R_ADMIN & X.holder.rights)) // we don't want to count admins twice. This list should be JUST mentors
+		if((R_MENTOR & X.holder.rights) && (!(R_ADMIN & X.holder.rights) || !(R_MOD & X.holder.rights))) // we don't want to count admins twice. This list should be JUST mentors
```
Rohesie:
```suggestion
		if((R_MENTOR & X.holder.rights) && !(R_ADMIN & X.holder.rights) && !(R_MOD & X.holder.rights)) // we don't want to count admins twice. This list should be JUST mentors
```
The negation of `A || B` is either `!A && !B` or `!(A || B)`, not `!A || !B`
Choose whichever you find more clear.

PR #585, opened 2018-10-28
Alamayer - A lowpop marine ship map

Alterist:
A test of the Alamayer, which is a tiny ship map that is missing most of the usual equipment. 
The dropship has been modified to include command equipment, a medbay and marine vendors. 
It's intended to be deployed mostly permanently, only going up to resupply. 
Latejoins spawn on the dropship, and can use the vendors there to gear up.
The dropship itself is surrounded with sentries, as losing the dropship will mean the loss of command


Alterist:
untitled-1

Ghommie:
The map is very WIP, and I'm sure you aknowledge that, so I'll leave some opinions anyway.

- CiC lacks a proper room for the armory, or a row closets with SO level access (afaik you can gun those open) at the corner. It's to keep the room tidy rather than restricting access.
- remove some of the random barrels/boxes cluttering the map to optimize the room space.
- The CO's bedroom should be at least 3 tiles vertical. Same kind of applies to CL's.
- The dropship mounted medical may seem fancy, but it will work horribly. Make a proper medbay at the dropship's ammunition room, relocate the latter and shrink down the relax/bedrooms area (one arcade machine suffices) to expand the former, so we don't witness a micro-station meme.
- The overwatch consoles should be located at the current security's checkpoint rather than on dropship, also it'd be good to expand it a bit.
- Is the elevator just below said checkpoint still needed? Can it be safely removed from the map?
- ARES should be somewhere less faraway, even a backroom of the CiC's restroom would be a better. 
- Is there really the need of that many generators? 6-9 should suffice imo.
- The out-of-cryo food service rooms shouldn't be as large since it's aimed for low-pop.

P.S. If in the future we make the dropship's more modular and less hardcoded, The second hangar optional and the dropships be able to choose which destination to fly at.

francinum:
I'd suggest toying with removing the job restrictions on the specialty vendors on the shuttle.

Ghommie:
Is there the possibility to make this into a different file instead of writing over the almayer's one, anyway?
It's be a good to preserve it.

PR #566, opened 2018-10-24
Actually fixes rogue autoinjectors

Ghommie:
Fixes #533.
Changes to some autoinjectors' paths to match pill bottles'.
Changes to autoinjectors and pills naming, removed Capitalization.
Removed excess icon updating, as it already updates on reagents change.
"expended" is now a pre-fix instead of suffix.
Renamed many istances of "quick clot"/"quickclot" to quick-clot for consistency.
Minor grammar things with injections and cqc security.

EDIT: I know it conflicts with #534, but we heard it didn't really work well. 
@Surrealaser, no need to close your PR enterely, just resolve it later since the dexalin plus portion is good.

RE-EDIT: Even more repathing/rewriting. Removed useless/unused files/paths.

PR #589, opened 2018-10-28
Fixes anti-tox change in #566

CosmicScientist:
#566 removed the declaration/definition of antitox:
https://github.com/ColonialMarines-Mirror/ColonialMarines-2018/pull/566/files#diff-ebcea53e6caa93556d0057343cb88e24L120

This resulted in an empty pill.


CosmicScientist:
Actually, slight problem, dylovene pill bottles use this pill instead of their own (defined right above it in the code).

CosmicScientist:
And actually doesn't work because a lot of anti-tox entries were changed... not actually sure what form anti-tox still exists in because the mentioned change PR is very bigly.

I'll correct this line in another web edit PR:
https://github.com/ColonialMarines-Mirror/ColonialMarines-2018/blob/master/code/game/objects/items/storage/firstaid.dm#L239

Issue #587, opened 2018-10-28
Dylovene pills are empty


Anonmare:
I'm not talking about the pill bottles, the _actual pills_ are empty, they have no reagents in them. According to Francinium, they have an undefined path.

CosmicScientist and Francinium can confirm.

CosmicScientist:
#589 half fixes this issue (offending PR mentioned there too). Dylovene pill bottles use anti-tox pills which are undeclared in the code whilst dylovene pills are fully functional if the reagent works as intended.

PR #591, opened 2018-10-29
Changes dylovene bottle to use dylovene

CosmicScientist:
Does not rename object name to pill_bottle/dylovene, so I'll see if I can do a find/replace before this is ready for a merge.


CosmicScientist:
Is a redo of #589 and therefore is to correct #566 and fix #587

Rohesie:
Being fixed in #594

PR #594, opened 2018-10-29
Further pill changes, have a medical candy.

### LINKED ISSUE #587: Dylovene pills are empty

Ghommie:
Fixes #587, fixes a similar problem with soporific pills. Conflicts with #591.

Plus deletes a contraband.dm containing only the zoom/happy pill bottles, moves them to firstaid.dm.

Ultrazine pill bottle is now a subtype of pill_bottle/restricted, boosted a little the list of possible requirements to open restricted pill_bottles (I should make this into a storage level thing and further improve it in future).

Made the get_idcard() into a mob/proc. It also checks for hands now. Made obj/proc/allowed() follow it.

Fixes a pre-existent problem with an invisible sprite when taking it out the bottle through attack_self() while looking the inside of the container. One hand bandits will now drop the pill on the floor instead of being blocked (I could make this into a skillcheck but meh).


code/game/objects/items/storage/firstaid.dm
```diff
@@ -304,45 +307,94 @@
...
+			return FALSE
 
 	return TRUE
 
-/obj/item/storage/pill_bottle/ultrazine/attack_self(mob/living/user)
-	if(!id_check(user))
+/obj/item/storage/pill_bottle/restricted/attack_self(mob/living/user)
+	if(!scan(user))
 		return
 	..()
```
Rohesie:
```suggestion
	return ..()
```

code/game/objects/items/storage/firstaid.dm
```diff
@@ -304,45 +307,94 @@
...
 
-/obj/item/storage/pill_bottle/ultrazine/open(mob/user)
-	if(!id_check(user))
+/obj/item/storage/pill_bottle/restricted/open(mob/user)
+	if(!scan(user))
 		return
 	..()
```
Rohesie:
```suggestion
	return ..()
```

PR #595, opened 2018-10-29
Husking Removed

Surrealaser:
Per title.


Ghommie:
You should either make this a config or also remove stuff related to it. Don't just remove the istances of a feature and leave remaining bits to stagnate the code.

Surrealaser:
> You should either make this a config or also remove stuff related to it. Don't just remove the istances of a feature and leave remaining bits to stagnate the code.

This only removes husking via fire overkill only. If admuns want to husk during an event or whatever, they still have that capability by design.

DominikPanic:
Fuck I'm retarded

Ghommie:
> 
> This only removes husking via fire overkill only. If admuns want to husk during an event or whatever, they still have that capability by design.

That's why I'd prefer this to be a config, nothing prevents fire-husking from being off by default. This is so admemes can turn it on during events or for "hardcore".
And I'm not even telling you to make an admin button for it.

Rohesie:
The request is fair.

code/__DEFINES/__game.dm
```diff
@@ -208,3 +208,6 @@ var/static/list/scarySounds = list('sound/weapons/thudswoosh.ogg','sound/weapons
 
 // Default preferences
 #define DEFAULT_SPECIES "Human"
+
+//Husking Switch For Burning
+#define HUSKING_ON 0
```
Ghommie: Make it configuration var on game_options.dm rather than a define, so it is actually toggleable.
Also `FALSE`.

code/modules/mob/living/carbon/human/human_damage.dm
```diff
@@ -17,8 +17,10 @@
 
 	health = species.total_health - oxy_l - tox_l - clone_l - total_burn - total_brute
 
+#if HUSKING_ON
```
Ghommie: `if(config.example_fire_husking_config_name_on)`

Surrealaser: This has been done.

code/modules/mob/living/carbon/human/human_damage.dm
```diff
@@ -17,8 +17,10 @@
 
 	health = species.total_health - oxy_l - tox_l - clone_l - total_burn - total_brute
 
-	if(((species.total_health - total_burn) < config.health_threshold_dead * 1.5))
-		ChangeToHusk()
+	if(config.husking_on)
```
Rohesie:
```suggestion
	if(config.husking_on && ((species.total_health - total_burn) < config.health_threshold_dead * 1.5))
```

code/modules/mob/living/carbon/human/human_damage.dm
```diff
@@ -17,8 +17,10 @@
...
+		if(((species.total_health - total_burn) < config.health_threshold_dead * 1.5))
```
Rohesie:
```suggestion
```

code/modules/mob/living/carbon/human/human_damage.dm
```diff
@@ -17,8 +17,10 @@
...
+			ChangeToHusk()
```
Rohesie:
```suggestion
		ChangeToHusk()
```

code/__DEFINES/__game.dm
Rohesie:
```suggestion
```

code/__DEFINES/__game.dm
```diff
@@ -208,3 +208,6 @@ var/static/list/scarySounds = list('sound/weapons/thudswoosh.ogg','sound/weapons
 
 // Default preferences
 #define DEFAULT_SPECIES "Human"
+
+//Husking Switch For Burning
```
Rohesie:
```suggestion
```

code/__DEFINES/__game.dm
```diff
@@ -208,3 +208,6 @@ var/static/list/scarySounds = list('sound/weapons/thudswoosh.ogg','sound/weapons
 
 // Default preferences
 #define DEFAULT_SPECIES "Human"
+
```
Rohesie:
```suggestion
```

Rohesie:
Still need to ask that option to roundstart config options to read it from the file, but other than that it's ready to go.

PR #608, opened 2018-10-31
Halloween Hall of Monsters

Alterist:
It's halloween! The xenos have decided to dress up to do some trick or treating on the marines


Ghommie:
Got a mix of feelings toward this, but I can just say it's just funny, specially the lying srpites being simply rotated sprites.
A little issue though is the drone using the same tiny sprite of the facehugger, that may really spook players. replace those with the vanilla spess man alien drone's.
Hopefully you'll get a speedy testmerge.

PR #610, opened 2018-11-01
Ravager Rework - Angry Red Dinosaur

Surrealaser:
-Ravager gains two new abilities and one reworked ability, as well as the Rage mechanic.

Rage: The Ravager gains 5 stacks of this whenever it makes a melee harm attack, or it takes damage, generating rage equal to 30% of the damage dealt to a maximum of 50 stacks total. Each point of Rage increases its movement and attack speed, fire resistance (it's now fire resistant instead of immune), recovery rate from staggerstun, stun and knockdown effects, and general damage resistance (max 60%). 

If Rage has not increased in the past 3 seconds, Rage will decrease automatically by 5 on each subsequent tick.

Rage is also used to enhance the damage output of its special abilities and is consumed by them.

Ravage: 10 second cooldown. The Ravager dashes forward up to one tile away and then makes a slash attack against up to three adjacent enemies it's facing that are then knocked back one tile. These attacks each deal bonus damage equal to 1% of the Ravager's base damage times its rage. The Ravager then loses all Rage, but for each victim, regains 10 Rage. Cooldown is reduced by 3 seconds per victim.

Eviscerating Charge: 30 second cooldown. This replaces its old Charge ability. Like the existing charge except that it moves further, and it deals +4% of the Ravager's base damage for each point of Rage, while sending its victim flying. This depletes Rage completely, and is best used as a finisher after having built up Rage, or to assassinate a key enemy, as it can nearly insta kill or badly fuck up at least most marines at or near maximum rage.

Second Wind: 60 second cooldown. Consume all Rage after a 5 second channel to regain health and plasma equal to 25% of the amount missing + 1.5% of this amount per point of Rage (total max 100% obviously). Cooldown is reduced by 1% per point of Rage.

-To adjust for these changes, Ravager armour has been reduced, its movement speed has been slightly reduced, it no longer suffers only minimal stun from explosives, and it has 50% base fire resistance instead of fire immunity.


NonQueueingMatt:
How severe is the armour reduction?

Supermichael777:
20 points. its kinda hard to contextualize but let me run the numbers
used to be 40/45/50/50
now 20/25/28/30
rifle bullets do 40 hit damage at ap 10
that means your effective armor is 10/15/18/20 formerly 30/35/40/40
Thats an over 50% effective nerf vs the common weapon

initial armor soak is 2/3/4/5  formerly 7/8/10/10
so damage(excluding crits) is 38/37/36/35 vs 33/32/30/30
and armor is  9/14/16/18  vs  27/31/35/35
then its damage halving/quartering time
armor_soak 	 = round(damage * 0.5) // floors it BTW always rounds down
armor 		-= armor_soak * 1.5
damage -= armor_soak
so
soak 19/18/18/17 vs 16/16/15/15  //24/24/22.5/22.5
armor 0!/0!/0!/0! vs  3/7/12.5/12.5
damage 19/19/18/18 vs 17/16/15/15
soak 0/0/0/0 vs 8/8/7/7
damage 19/19/18/18 vs 9/8/8/8

so yeah that's fairly substantial on average doubling damage

Surrealaser:
@Supermichael777 That's why you gain general damage resistance with rage, which reduces all damage by up to 60%.

Rohesie:
This makes it less of a hit-and-run xeno, but more of a glass canon. Interesting.

My only concern by reading the description is the cooldown reduction on some cases. It looks like it will be able to Ravage once per second as long as it has targets in sight. Could be good, or could be bad...

code/__DEFINES/mobs.dm
```diff
@@ -458,3 +458,13 @@ var/list/global_mutations = list() // list of hidden mutation things
 #define MOTH_WINGS_BEHIND_LAYER	1
 
 #define TOTAL_UNDERLAYS			1
+
+
+//Ravager defines:
+#define CHARGESPEED					3
+#define CHARGESTRENGTH				3
+#define CHARGEDISTANCE				5
+#define CHARGECOOLDOWN				300
```
Rohesie: Charge is pretty generic so you'll probably do better by adding RAV_ at the start.

code/modules/mob/living/carbon/xenomorph/Castes/Ravager.dm
```diff
@@ -14,33 +14,31 @@
...
-	var/CHARGECOOLDOWN = 120
 	charge_type = 3 //Claw at end of charge
-	fire_immune = 1
-	armor_deflection = 40
+	fire_resist = 0.5
+	armor_deflection = 20
 	mob_size = MOB_SIZE_BIG
 	drag_delay = 6 //pulling a big dead xeno is hard
-	xeno_explosion_resistance = 2 //no stuns from explosions
+	xeno_explosion_resistance = 1 //can't be gibbed from explosions
```
Rohesie: This could really use a define.

code/modules/mob/living/carbon/xenomorph/Castes/Ravager.dm
```diff
@@ -14,33 +14,31 @@
...
 	attack_delay = -2
 	tier = 3
 	upgrade = 0
 	pixel_x = -16
 	old_x = -16
+	var/usedcharge = 0 //What's the deal with the all caps?? They're not constants :|
```
Rohesie:
```suggestion
	var/usedcharge = FALSE //What's the deal with the all caps?? They're not constants :|
```

code/modules/mob/living/carbon/xenomorph/Castes/Ravager.dm
```diff
@@ -64,11 +63,16 @@
...
 	usedPounce = 1 //This has to come before throw_at, which checks impact. So we don't do end-charge specials when thrown
-	use_plasma(20)
+	use_plasma(80)
+
 	throw_at(T, CHARGEDISTANCE, CHARGESPEED, src)
+
+	charge_delay = world.time + CHARGECOOLDOWN
+
 	spawn(CHARGECOOLDOWN)
 		usedPounce = 0
```
Rohesie:
```suggestion
		usedPounce = FALSE
```

code/modules/mob/living/carbon/xenomorph/Powers.dm
```diff
@@ -1462,3 +1463,105 @@
...
+	var/list/L = orange(sweep_range)		// Not actually the fruit
+	var/victims
+	var/target_facing
+	for (var/mob/living/carbon/human/H in L)
+		if(victims >= 3) //Max 3 victims
+			break
+		target_facing = get_dir(src, H)
+		if(target_facing != dir && target_facing != turn(dir,45) && target_facing != turn(dir,-45) ) //Have to be actually facing the target
+			continue
+		if(H.stat != DEAD && !(istype(H.buckled, /obj/structure/bed/nest) && H.status_flags & XENO_HOST) ) //No bully
```
Rohesie: Early continue?

code/modules/mob/living/carbon/xenomorph/XenoProcs.dm
```diff
@@ -145,6 +145,9 @@
 	if(frenzy_aura)
 		. -= (frenzy_aura * 0.05)
 
+	if(rage)
```
Rohesie: You can put this into `/mob/living/carbon/Xenomorph/Ravager/movement_delay()` and avoid ravager vars for non-Ravas, as well as redundant checks.
Don't forget to add a `. = ..()` at the beginning.

code/modules/mob/living/carbon/xenomorph/XenoProcs.dm
```diff
@@ -262,17 +265,20 @@
...
+					M.KnockDown(1)
 				if(4) //Predalien.
 					M.attack_alien(src) //Free hit/grab/tackle. Does not weaken, and it's just a regular slash if they choose to do that.
 
 		throwing = FALSE //Resert throwing since something was hit.
 		return TRUE
-
+	if(charge_type == RAV_CHARGE_TYPE) //Hit or miss, after we apply bonuses from charge, rage is 0ed out.
+		rage = 0
 	..() //Do the parent otherwise, for turfs.
```
Rohesie:
```suggestion
	return ..() //Do the parent otherwise, for turfs.
```

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -29,6 +29,19 @@
...
+		if(world.time > last_rage + 30) //Decrement Rage if it's been more than 3 seconds since we last raged.
+			rage = CLAMP(rage - 5,0,50) //Rage declines over time.
+		AdjustStunned( round(-rage * 0.1,0.01) ) //Recover 0.1 more stun stacks per unit of rage; min 0.1, max 5
+		AdjustKnockeddown( round(-rage * 0.1, 0.01 ) ) //Recover 0.1 more knockdown stacks per unit of rage; min 0.1, max 5
+		adjust_slowdown( round(-rage * 0.1,0.01) ) //Recover 0.1 more stagger stacks per unit of rage; min 0.1, max 5
+		adjust_stagger( round(-rage * 0.1,0.01) ) //Recover 0.1 more stagger stacks per unit of rage; min 0.1, max 5
+		rage_resist = CLAMP(1-round(rage * 0.012,0.01),0.4,1) //+1.2% damage resist per point of rage
+		fire_resist = initial(fire_resist) - round(rage * 0.01,0.01) //+1% fire resistance per stack of rage, max +50%; initial resist is 50%
+		attack_delay = initial(attack_delay) - round(rage * 0.05,0.01) //-0.05 attack delay to a maximum reduction of -2.5
+	. = ..()
```
Rohesie:
```suggestion
	return ..()
```
If it's at the end, return instead.
This all stacks with other reductions, as you probably know.

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -91,8 +104,9 @@
 /mob/living/carbon/Xenomorph/handle_fire()
 	if(..())
```
Rohesie:
```suggestion
	. = ..()
	if(.)
```

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
```diff
@@ -159,6 +160,16 @@
...
 
+	//Ravager vars
+	var/rage = 0
+	var/rage_resist = 1.00
+	var/ravage_used = FALSE
+	var/ravage_delay = null
+	var/charge_delay = null
+	var/second_wind_used = FALSE
+	var/second_wind_delay = null
+	var/last_rage = null
```
Rohesie: These should be ideally in the Ravager definition.

Supermichael777:
Fun fact raw damage is so good that shotguns arent even more effective against a 50% change

code/modules/mob/living/carbon/xenomorph/Powers.dm
```diff
@@ -1467,8 +1467,106 @@
 	spawn(savage_cooldown)
 		savage_used = FALSE
 		to_chat(src, "<span class='xenowarning'><b>You can now savage your victims again.</b></span>")
-		playsound(src, "xeno_newlarva", 100, 0, 1)
-		update_action_button_icons()
```
Rohesie: This should be preserved.

code/modules/mob/living/carbon/xenomorph/XenoProcs.dm
```diff
@@ -264,18 +271,31 @@
...
 					M.attack_alien(src) //Free hit/grab/tackle. Does not weaken, and it's just a regular slash if they choose to do that.
 
 		throwing = FALSE //Resert throwing since something was hit.
 		return TRUE
+	process_ravager_charge(FALSE)
+	return ..() //Do the parent otherwise, for turfs.
 
-	..() //Do the parent otherwise, for turfs.
+/mob/living/carbon/Xenomorph/proc/process_ravager_charge(hit = TRUE, mob/living/carbon/M = null)
+	if(!istype(src,/mob/living/carbon/Xenomorph/Ravager))
```
Rohesie: Make this `return FALSE` and redefine this for `/mob/living/carbon/Xenomorph/Ravager/process_ravager_charge(hit = TRUE, mob/living/carbon/M = null)`

code/modules/mob/living/carbon/xenomorph/attack_alien.dm
```diff
@@ -205,6 +219,14 @@
 	return TRUE
 
 
+/mob/living/carbon/Xenomorph/proc/process_rage_attack()
+	if(!istype(src,/mob/living/carbon/Xenomorph/Ravager))
```
Rohesie: `return FALSE` and redefine for `/mob/living/carbon/Xenomorph/Ravager/process_rage_attack()`

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
```diff
@@ -118,6 +120,15 @@
 /mob/living/carbon/Xenomorph/adjustFireLoss(amount)
 	fireloss = CLAMP(fireloss + amount, 0, maxHealth - crit_health)
 
+/mob/living/carbon/Xenomorph/proc/process_rage_damage(damage)
+	if(!damage || !istype(src,/mob/living/carbon/Xenomorph/Ravager))
```
Rohesie: `return damage` (if that makes sense) and redefine for `/mob/living/carbon/Xenomorph/Ravager/process_rage_damage(damage)`

PR #612, opened 2018-11-02
2 Walls 2 Furious (Bay walls round 2)

Arachnidnexus:
Since there a ton of walls, I will be using a better naming scheme for the PRs now. This one focuses on normal metal walls and reinforced metal walls and associated windows and window frames.

-metal walls/windows/window frames changed to ones based off Bay walls
image
-reinforced walls/windows/window frames using metal walls as base
image
-fixes ship hull windows reverting to wrong window frame
image
-cleans up ship medbay windows (North/South) a little
image


Rohesie:
Beautiful.

PR #617, opened 2018-11-02
Xeno Queen rolling rework + monkey rebalance.

DominikPanic:
-Queen is now rolled before xeno from everyone who has queen on, instead of just the available xenos.
-Amount of starting monkeys reduced to 5, scales with player population like the requisition vendors.

### DIFF
--- a/code/game/gamemodes/cm_initialize.dm
+++ b/code/game/gamemodes/cm_initialize.dm
@@ -90,7 +90,8 @@ Additional game mode variables.
 
 datum/game_mode/proc/initialize_special_clamps()
 	var/ready_players = num_players() // Get all players that have "Ready" selected
-	xeno_starting_num = max((ready_players/7), xeno_required_num) //(n, minimum, maximum)
+	//xeno_starting_num = max((ready_players/7), xeno_required_num) //(n, minimum, maximum)
+	xeno_starting_num = 1
 	surv_starting_num = CLAMP((ready_players/25), 0, 8)
 	merc_starting_num = max((ready_players/3), 1)
 	marine_starting_num = ready_players - xeno_starting_num - surv_starting_num - merc_starting_num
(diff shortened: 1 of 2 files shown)


code/game/gamemodes/cm_initialize.dm
```diff
@@ -312,9 +313,41 @@ datum/game_mode/proc/initialize_special_clamps()
...
 
+/datum/game_mode/proc/initialize_starting_queen_list()
+	var/list/datum/mind/possible_queens = get_players_for_role(BE_QUEEN)
+
+	//Minds are not transferred at this point, so we have to clean out those who may be already picked to play.
+	for(var/datum/mind/A in possible_queens)
+		if(A.assigned_role == "MODE")
+			possible_queens -= A
+
+	if(!possible_queens.len)
```
Rohesie:
```suggestion
	if(!length(possible_queens))
```
Oddly, better performance.

code/game/gamemodes/cm_initialize.dm
```diff
@@ -292,11 +292,12 @@ datum/game_mode/proc/initialize_special_clamps()
 	while(i > 0) //While we can still pick someone for the role.
 		if(possible_xenomorphs.len) //We still have candidates
```
Rohesie:
```suggestion
		if(length(possible_xenomorphs)) //We still have candidates
```

code/game/gamemodes/cm_initialize.dm
```diff
@@ -484,9 +527,11 @@ datum/game_mode/proc/initialize_special_clamps()
 			var/i = surv_starting_num
 			var/datum/mind/new_survivor
 			while(i > 0)
-				if(!possible_survivors.len) break  //Ran out of candidates! Can't have a null pick(), so just stick with what we have.
+				if(!possible_survivors.len)
```
Rohesie:
```suggestion
				if(!length(possible_survivors)) 
```

PR #620, opened 2018-11-03
Bipods made more mobile, allows use of window frames as support

Jalordon:
(based on #489 , fumbled with git and some reason it unlinked from my branch)

Changes to bipods:
- can always be deployed, even without valid support. 
- makes intended slowdown apply immediately when bipod is deployed, instead of only when re-wielded
- no longer auto-retracts without support, so its less management of the bipod if accidentally shooting at an angle without cover
- will now accept window frames on the tile infront of the shooter as valid support, giving some utility away from forward bases.

Changes intended to make bipods less effort to use and applicable without barricades, hopefully making them a more viable choice compared to other under barrel attachments.


code/modules/projectiles/updated_projectiles/gun_system.dm
```diff
@@ -624,7 +631,7 @@ and you're good to go.
 		var/scatter_chance_mod = 0
 		var/burst_scatter_chance_mod = 0
 		//They decrease scatter chance and increase accuracy a tad. Can also increase damage.
-		if(user && under && under.bipod_deployed) //Let's get to work on the bipod. I'm not really concerned if they are the same person as the previous user. It doesn't matter.
+		if(flags_item & WIELDED && user && under && under.bipod_deployed) //Let's get to work on the bipod. I'm not really concerned if they are the same person as the previous user. It doesn't matter.
```
Rohesie:
```suggestion
		if(flags_item & WIELDED && user && under?.bipod_deployed) //Let's get to work on the bipod. I'm not really concerned if they are the same person as the previous user. It doesn't matter.
```

PR #629, opened 2018-11-05
Walls 3: Shuttle Drift

Arachnidnexus:
I debated doing Chigusa Drift but those walls aren't even being used right now. Anyway, this is about yet another 3/4ths walls thing. This time for the ~~escape pods and~~ ERT shuttles.

_Edit: _ it-has-come-to-the-point-...-very-tired-...-walls.jpg

_Edit 2:_ fixed misaligned pixels but too lazy to upload new pictures. pretty much done with this for now unless people have objections to the inner walls or something

_Edit 3:_ reverting escape pods back to original ones since those apparently did not need to be replaced

generic ert:
image

corporate ert (NT used as placeholder, can be pretty easily changed):
image

human antag faction ert (COM as placeholder, using current faction colors):
image

to-do:
maybe make walls less bland, window more window-y


Alterist:
Shuttles are great, but in my opinion, the Evac pod's colors are too garish and doesn't match the rest of the ship. perhaps there's a better shade for them?

Arachnidnexus:
Removed the escape pods until they're reworked for something else maybe
_new orange creamsicle escape pods:_
image

PR #630, opened 2018-11-05
New sentry sprite (sprites only no code)

Arachnidnexus:
New sentry sprite made from scratch and put under **icons\Marine\new_sentry.dmi**. It has the following states:

sentry base (no battery, no ammo, not active)
active overlay
tipped over sentry
sentry turret head
battery overlays (green, yellow, orange, red, black)
ammo overlays (has ammo, no ammo)
tripod (deployed, folded)
sentry wreck

_edit 2: alternative new palette_
image
stored in **new_sentry_alt.dmi** in same folder

_edit 1: new palette_
image

_edit 1: old palette_
image

Would need someone to code this in because it uses a different sprite setup than the current sentry.


Ghommie:
I think it should need a catchier color palette, less light shading and a softer tonality, to better distinguish each of its components, directions and states. Also you misplaced the front leg behind the tripod's axis on its north sprite. Good job overrall.

Arachnidnexus:
@Ghommie I wasn't sure how to represent the front tripod being under and in front of the connector on the North sprite but yeah it looks a bit off. Gonna try to fix that and pick a different color palette.

Ghommie:
By catchier palette I meant a wider selection of colors to be used on the sprites, to differentiate the components of the gun like laser sight etc. I was pretty vague, my fault.
About the tripod front leg, you just got to move it a few tiles above, about where the other two junction.
I like the alternative palette and its dark shadows, they define its shape better.

PR #631, opened 2018-11-05
New Carrier Ability: Spawn Hugger

Surrealaser:
-Carriers now have the Spawn Hugger ability, allowing them to increment their facehugger stores by +1 at the cost of a 10 second cooldown and 100 plasma.

-This should make multiple Carriers more viable, and allow them to be relatively independent rather than completely and wholly reliant on Ovi'ed Queens.


Supermichael777:
This makes it so you don't have to risk a forward ovi or have a drone train of eggs to add hugers to combat.  Combat hugs are a ridiculous stun, especially because CM's nerfs and removals for it also removed marine huger counter-play(shakeup and self knife or stab in the face to "save" your friend).

Carriers had regenerating hugers for the longest time and it was just the worst.  Used to they just utterly trashcaned marine assaults because they massively increased stun volume on xeno flanks.

Its worth testing but i think it will break the hive balance.

Rohesie:
Combat hugging is nerfed, though, and so is the stun time.

Anonmare:
Hugger stun times were halved and don't remove helmets anymore

PR #634, opened 2018-11-06
HOTFIX: Corrected Tackle Halloss

Surrealaser:
It looks like a 2x multiplier was deleted at some point for the gaussian I used to calculate variance on halloss damage. Corrected this by just doubling the multipliers.


code/modules/mob/living/carbon/xenomorph/XenoUpgrade.dm
```diff
@@ -456,7 +456,6 @@ Queen		 0.0	 0.1	 0.2	 0.3
 					tacklemax = 5
 					tackle_damage = 40 // Prior was 30
 					pounce_delay = 30
-				if("Hunter")
```
Rohesie: Is this intended?

Surrealaser: Sec

PR #637, opened 2018-11-07
Unborks blindness.

### LINKED ISSUE #609: Blindness from blowtorch
Well apparently when you are supposed to go blind from blowtorch it does not blind you. Well it does blind you but the blindness overlay does not appear at all. However you cant climb ladders or anything like that while being blind.

I think that the error is in overlay and thats about it. I've had health hud on as well so that might have bugged it out. And I was repairing the synth's dented parts.

Ghommie:
Should close #609, makes ladders climbable by blind fellows (still can't look up and down). Makes the welder/plasmacutter eyecheck less crap for now (Will eventually substitute it with flash_act() on another PR anyway).
Fiddles with ear damage/deaf procs.
Makes carbon's update_stat() not run blindness procs more than needed, also preventing a faster un-blinding by repeatedly invoking it through updatehealth().

### DIFF
--- a/code/modules/mob/mob_helpers.dm
+++ b/code/modules/mob/mob_helpers.dm
@@ -102,7 +102,7 @@ proc/isdeaf(A)
 	if(isliving(A))
 		var/mob/living/M = A
 		return M.ear_deaf
-	return 0
+	return FALSE
 
 proc/is_blind(A)
 	if(isliving(A))


code/modules/mob/mob_helpers.dm
```diff
@@ -99,15 +99,15 @@ proc/isorgan(A)
 	return 0
 
 proc/isdeaf(A)
-	if(istype(A, /mob))
-		var/mob/M = A
-		return (M.sdisabilities & DEAF) || M.ear_deaf
+	if(isliving(A))
+		var/mob/living/M = A
+		return M.ear_deaf
 	return 0
```
Rohesie:
```suggestion
	return FALSE
```

PR #644, opened 2018-11-07
Return to Lobby using Hypersleep.

ActualRory:
Vote

Implements the above.

You can now return to the lobby using Hypersleep Cryopods.

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -9,7 +9,6 @@
 //Used for logging people entering cryosleep and important items they are carrying.
 var/global/list/frozen_crew = list()
 var/global/list/frozen_items = list("Alpha"=list(),"Bravo"=list(),"Charlie"=list(),"Delta"=list(),"MP"=list(),"REQ"=list(),"Eng"=list(),"Med"=list())
-var/num_respawn = 0
 
 //Main cryopod console.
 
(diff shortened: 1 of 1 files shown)


Ghommie:
Sounds ok, but shouldn't a "limit" to how many times a given player can return to the lobby be coded there?
Just to limit the aforementioned cons of this function.

ActualRory:
> Sounds ok, but shouldn't a "limit" to how many times a given player can return to the lobby be coded there?
> Just to limit the aforementioned cons of this function. Probably could have other requirements in future though.

This is a good idea. I will add this.

code/game/machinery/cryopod.dm
```diff
@@ -187,229 +187,231 @@ var/global/list/frozen_items = list("Alpha"=list(),"Bravo"=list(),"Charlie"=list
...
+	announce.autosay("[occupant.real_name] has entered long-term hypersleep storage. Belongings moved to hypersleep inventory.", "Hypersleep Storage System")
+	visible_message("<span class='notice'>[src] hums and hisses as it moves [occupant.real_name] into hypersleep storage.</span>")
 
-			cdel(occupant)
-			occupant = null
-			stop_processing()
+	//Delete the mob.
+	var/mob/new_player/NP = new()
+	NP.ckey = occupant.ckey
+	if(NP.client) NP.client.change_view(world.view)
```
Rohesie:
```suggestion
	NP.client?.change_view(world.view)
```

Rohesie: Remember not to use single-line control statements in the future, as well.

code/game/machinery/cryopod.dm
```diff
@@ -472,6 +474,23 @@ var/global/list/frozen_items = list("Alpha"=list(),"Bravo"=list(),"Charlie"=list
...
 
+/obj/machinery/cryopod/verb/return_to_lobby()
+	set name = "Return to Lobby"
+	set category = "Object"
+	set src in oview(1)
+	if(occupant != usr)
+		to_chat(usr, "<span class='warning'>You must be inside the [src] to do this.</span>")
+		return
+
+	if(alert(usr, "Do you want to send yourself back to the lobby? This will be logged.","Message","Yes","No") == "Yes")
```
Rohesie: Do a `== "No"` then `return` to diminish indentation. We call that an early return.

code/game/machinery/cryopod.dm
```diff
@@ -472,6 +474,23 @@ var/global/list/frozen_items = list("Alpha"=list(),"Bravo"=list(),"Charlie"=list
...
+		log_admin("[usr.mind.assigned_role] [key_name(usr)] returned to the lobby using a cryopod.")
+		message_admins("\blue [usr.mind.assigned_role] [key_name(usr)] returned to the lobby using a cryopod.")
+		var/mob/new_player/NP = new()
+		NP.ckey = occupant.ckey
+		if(NP.client) NP.client.change_view(world.view)
```
Rohesie:
```suggestion
		NP.client?.change_view(world.view)
```

code/game/machinery/cryopod.dm
Rohesie: After the alert is called the user may no longer be inside the cryo machine. That's something to consider.

Rohesie (changes requested):

code/game/machinery/cryopod.dm
ActualRory: Thanks for the tip!

code/game/machinery/cryopod.dm
ActualRory: Turns out the return was redundant and not needed at all? Go figure.

ActualRory:
Added respawn tracking and logging.
Hard limit is now three returns.
Soft limit is when admin raises an eyebrow at the notice.

Rohesie:
Make the respawn a client var, and place only one respawn by default, please.

ActualRory:
Will do!

code/game/machinery/cryopod.dm
```diff
@@ -472,6 +474,28 @@ var/global/list/frozen_items = list("Alpha"=list(),"Bravo"=list(),"Charlie"=list
 			//Despawning occurs when process() is called with an occupant without a client.
 			add_fingerprint(M)
 
+/obj/machinery/cryopod/verb/return_to_lobby()
+	set name = "Return to Lobby"
+	set category = "Object"
+	set src in oview(1)
```
Rohesie:
```suggestion
	set src in view(0)
```

Rohesie: `oview(1)` does a 3x3 scan of the object excluding the center.
`view(0)` only includes the center, which is what we want, for this to appear when the user is in the same tile.

code/game/machinery/cryopod.dm
```diff
@@ -472,6 +474,28 @@ var/global/list/frozen_items = list("Alpha"=list(),"Bravo"=list(),"Charlie"=list
...
+	if(occupant != usr)
```
Rohesie: Typecast `usr` here, please. Do we want to send back anything besides humans?
You are running some checks there like the mind one that is not necessarily defined for every possible `usr`.

Rohesie (changes requested):

PR #645, opened 2018-11-08
Corp Mercs Style SL Pointer/Locator

Surrealaser:
Also the existing SL pointer became much more responsive/less retarded.


code/modules/mob/living/carbon/human/human_movement.dm
```diff
@@ -80,6 +80,38 @@
...
 
+	Process_SL_Locator() //Adjust the angle of the overlay
+
+
+/mob/living/carbon/human/proc/Process_SL_Locator()
+	if(!sl_headset_active)
+		return
+	if(!assigned_squad)
+		return
+	if(world.time < sl_locator_next_update)
```
Rohesie: I'd place this check first. It would probably cover the other two cases since `sl_locator_next_update` wouldn't be updated anyway.

code/game/objects/items/devices/radio/headset.dm
```diff
@@ -259,15 +267,9 @@
 		if(src == user.wear_ear) //worn
 			var/datum/mob_hud/H = huds[MOB_HUD_SQUAD]
 			if(headset_hud_on)
-				H.add_hud_to(usr)
-				if(user.mind && user.assigned_squad && user.hud_used && user.hud_used.locate_leader)
-					user.hud_used.locate_leader.alpha = 255
-					user.hud_used.locate_leader.mouse_opacity = 1
+				add_squadhud(H, user)
```
Rohesie: This should only be added for mobs with assigned squads. Else the doctors, SOs and whatever will be calling this proc all the time, wasting resources.

code/modules/mob/living/carbon/human/human_movement.dm
```diff
@@ -80,6 +80,38 @@
...
+		return
+	var/mob/living/carbon/human/H = assigned_squad.squad_leader
+	var/obj/screen/SL = hud_used.SL_locator
+	var/obj/screen/SL_dial = hud_used.locate_leader
```
Rohesie: These vars are created here and destroyed once the proc ends, is that right?

Rohesie (changes requested):

code/game/objects/items/devices/radio/headset.dm
Rohesie: Though ideally there's a separate verb to toggle the tracker as well.

Rohesie: The `else` will now prevent from shipside roles from having a squad HUD.
Make these two different checks, please, and different verbs to toggle.

Rohesie (changes requested):

Rohesie (changes requested):

Rohesie (changes requested):

PR #646, opened 2018-11-08
LTB Cannon Now Has 0.5 Second Delay

Surrealaser:
-This should balance and allow for the lethality of the LTB cannon round, especially its direct hits.


Supermichael777:
Fire delays are cancer and lead to salt all around

code/modules/vehicles/hardpoints.dm
```diff
@@ -171,14 +171,15 @@ Currently only has the tank hardpoints
...
 
-		next_use = world.time + owner.cooldowns["primary"] * owner.misc_ratios["prim_cool"]
-		if(!prob(owner.accuracies["primary"] * 100 * owner.misc_ratios["prim_acc"]))
-			T = get_step(T, pick(cardinal))
-		var/obj/item/projectile/P = new
-		P.generate_bullet(new ammo.default_ammo)
-		P.fire_at(T, owner, src, P.ammo.max_range, P.ammo.shell_speed)
-		playsound(get_turf(src), pick('sound/weapons/tank_cannon_fire1.ogg', 'sound/weapons/tank_cannon_fire2.ogg'), 60, 1)
-		ammo.current_rounds--
+		if(do_after(usr, 5, FALSE, 5, BUSY_ICON_HOSTILE))
```
Rohesie: Early return instead?

Rohesie:
I'd like to see how this plays out a bit before merging.

PR #653, opened 2018-11-10
Scout Specialist Overhaul

Surrealaser:
-The thermal cape's cloaking now features a depleteable energy counter instead of a  fixed duration; when energy reaches 0, the cloaking ends.

-Cloaking initially grants 95% transparency for a short duration. This transparency is reduced to 80% while walking after a short delay, or 50% while running. Remaining still for 3 seconds will increase the transparency to 95%. 

-Energy is consumed at a rate of 1 per tile walked, and 5 per tile ran. Energy is slowly restored at a rate of 5 once every every 2 seconds roughly while stationary.

-Most aggressive actions like melee attacks, throwing and shooting will cause the Cloak to 'shimmer' capping transparency at 50% for a short duration. While shimmering, energy consumption for movement is maximized.

-M4RA rifle substantially improved; it now has more accuracy, less scatter, a slightly faster burst and can equip many more attachments like the barrel charger, extended barrel, laser sight, underslung shotgun and flamer.

### DIFF
--- a/code/__DEFINES/objects.dm
+++ b/code/__DEFINES/objects.dm
@@ -158,3 +158,20 @@ var/list/RESTRICTED_CAMERA_NETWORKS = list( //Those networks can only be accesse
 #define SENTRY_ALERT_BATTERY			5
 #define SENTRY_ALERT_DELAY				200 //20 seconds
 #define SENTRY_DAMAGE_ALERT_DELAY		50 //5 seconds
+
+
+//Scout cloak defines
+#define SCOUT_CLOAK_ENERGY	100
+#define SCOUT_CLOAK_STEALTH_DELAY 30
+#define SCOUT_CLOAK_RUN_DRAIN	5
+#define SCOUT_CLOAK_WALK_DRAIN	1
+#define SCOUT_CLOAK_ACTIVE_RECOVERY -5 //You only get this once every obj tick, so it'll be comparable to the inactive value
+#define SCOUT_CLOAK_INACTIVE_RECOVERY 5
+#define SCOUT_CLOAK_COOLDOWN 100
+#define SCOUT_CLOAK_TIMER 50
+#define SCOUT_CLOAK_RUN_ALPHA 128
+#define SCOUT_CLOAK_WALK_ALPHA 51
+#define SCOUT_CLOAK_STILL_ALPHA 13
+#define SCOUT_CLOAK_MAX_ENERGY 100
+#define SCOUT_CLOAK_OFF_DAMAGE 1
+#define SCOUT_CLOAK_OFF_ATTACK 2
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -267,6 +267,10 @@
 		src.thrower = null
 		src.throw_source = null
 
+	if(ishuman(thrower)) //Human specific procs
+		var/mob/living/carbon/human/H = thrower
+		H.camo_off_process(SCOUT_CLOAK_OFF_ATTACK)
+
 
 //Overlays
 /atom/movable/overlay
--- a/code/game/objects/items/storage/backpack.dm
+++ b/code/game/objects/items/storage/backpack.dm
@@ -409,18 +409,7 @@
 	icon_state = "smock"
 	worn_accessible = TRUE
 
-#define SCOUT_CLOAK_ENERGY	100
-#define SCOUT_CLOAK_STEALTH_DELAY 3
-#define SCOUT_CLOAK_RUN_DRAIN	5
-#define SCOUT_CLOAK_WALK_DRAIN	1
-#define SCOUT_CLOAK_ACTIVE_RECOVERY -5
-#define SCOUT_CLOAK_INACTIVE_RECOVERY -10
-#define SCOUT_CLOAK_COOLDOWN 100
-#define SCOUT_CLOAK_TIMER 50
-#define SCOUT_CLOAK_RUN_ALPHA 128
-#define SCOUT_CLOAK_WALK_ALPHA 205
-#define SCOUT_CLOAK_STILL_ALPHA 243
-#define SCOUT_CLOAK_MAX_ENERGY 100
+
 // Scout Cloak
 /obj/item/storage/backpack/marine/satchel/scout_cloak
 	name = "\improper M68 Thermal Cloak"
@@ -432,22 +421,31 @@
 	var/camo_active_timer = 0
 	var/camo_cooldown_timer = null
 	var/camo_last_stealth = null
+	var/camo_last_shimmer = null
 	var/camo_energy = 100
+	var/mob/living/wearer = null
+	actions_types = list(/datum/action/item_action/toggle)
 
 /obj/item/storage/backpack/marine/satchel/scout_cloak/Dispose()
+	camo_off()
+	wearer = null
 	processing_objects.Remove(src)
 	. = ..()
 
 /obj/item/storage/backpack/marine/satchel/scout_cloak/dropped(mob/user)
 	camo_off(user)
+	wearer = null
 	processing_objects.Remove(src)
 	. = ..()
 
-/obj/item/storage/backpack/marine/satchel/scout_cloak/verb/camouflage()
+/obj/item/storage/backpack/marine/satchel/scout_cloak/verb/use_camouflage()
 	set name = "Toggle M68 Thermal Camouflage"
 	set desc = "Activate your cloak's camouflage."
 	set category = "Scout"
 
+	camouflage()
+
+/obj/item/storage/backpack/marine/satchel/scout_cloak/proc/camouflage()
 	if (!usr || usr.is_mob_incapacitated(TRUE))
 		return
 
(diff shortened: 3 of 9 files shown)


code/__DEFINES/objects.dm
```diff
@@ -157,4 +157,21 @@ var/list/RESTRICTED_CAMERA_NETWORKS = list( //Those networks can only be accesse
...
\ No newline at end of file
+#define SENTRY_DAMAGE_ALERT_DELAY		50 //5 seconds
+
+
+//Scout cloak defines
+#define SCOUT_CLOAK_ENERGY	100
+#define SCOUT_CLOAK_STEALTH_DELAY 30
+#define SCOUT_CLOAK_RUN_DRAIN	5
+#define SCOUT_CLOAK_WALK_DRAIN	1
+#define SCOUT_CLOAK_ACTIVE_RECOVERY -5 //You only get this once every obj tick, so it'll be comparable to the inactive value
```
spookydonut: I'd flip this to use positive values imo.

code/__DEFINES/objects.dm
```diff
@@ -157,4 +157,21 @@ var/list/RESTRICTED_CAMERA_NETWORKS = list( //Those networks can only be accesse
...
+#define SCOUT_CLOAK_INACTIVE_RECOVERY 5
+#define SCOUT_CLOAK_COOLDOWN 100
+#define SCOUT_CLOAK_TIMER 50
+#define SCOUT_CLOAK_RUN_ALPHA 128
+#define SCOUT_CLOAK_WALK_ALPHA 51
+#define SCOUT_CLOAK_STILL_ALPHA 13
+#define SCOUT_CLOAK_MAX_ENERGY 100
+#define SCOUT_CLOAK_OFF_DAMAGE 1
```
spookydonut: Separate these bitflags into a headed section and give them a different name to avoid confusion.

code/game/atoms_movable.dm
```diff
@@ -267,6 +267,10 @@
 		src.thrower = null
 		src.throw_source = null
 
+	if(ishuman(thrower)) //Human specific procs
+		var/mob/living/carbon/human/H = thrower
```
spookydonut: Why bother casting this to human if you defined the proc as mob/living

code/game/atoms_movable.dm
```diff
@@ -267,6 +267,10 @@
...
+		H.camo_off_process(SCOUT_CLOAK_OFF_ATTACK)
```
spookydonut: having this called every time a human throws something feels like a bad idea.

code/game/objects/items/storage/backpack.dm
```diff
@@ -420,14 +419,33 @@
 	has_gamemode_skin = FALSE //same sprite for all gamemode.
 	var/camo_active = 0
 	var/camo_active_timer = 0
-	var/camo_cooldown_timer = 0
-	var/camo_ready = 1
+	var/camo_cooldown_timer = null
+	var/camo_last_stealth = null
+	var/camo_last_shimmer = null
+	var/camo_energy = 100
+	var/mob/living/wearer = null
```
spookydonut: unless `/mob/living/silicon`, `/mob/living/carbon/monkey`, etc are going to be using this you should use `var/mob/living/carbon/human/wearer`
Synths are `/mob/living/carbon/human`

code/modules/mob/living/carbon/human/human_damage.dm
```diff
@@ -261,6 +261,12 @@ In most cases it makes more sense to use apply_damage() instead! And make sure t
 	var/datum/limb/picked = pick(parts)
 	if(picked.take_damage(brute,burn,sharp,edge))
 		UpdateDamageIcon()
+
+	if(brute)
+		camo_off_process(SCOUT_CLOAK_OFF_DAMAGE, brute) //If we have the Scout cloak, check for a short out
```
spookydonut: same thing as above this proc will get called anytime humans take damage.

code/modules/mob/living/carbon/human/human_movement.dm
```diff
@@ -78,9 +78,32 @@
 	if(mRun in mutations)
 		. = 0
 
+	Process_Cloaking(src)
+
 	. += config.human_delay
 
 
+/mob/living/proc/Process_Cloaking(mob/living/carbon/human/user)
```
spookydonut: this is icky.

code/game/atoms_movable.dm
Rohesie: Use redefined procs instead of type checks whenever possible.

code/game/atoms_movable.dm
Rohesie: Agreed, unless if the proc is really cheap. Most cloaking procs are pretty intrusive this way.
Suggestions on alternatives?

code/game/objects/items/storage/backpack.dm
```diff
@@ -420,14 +419,33 @@
...
+	actions_types = list(/datum/action/item_action/toggle)
+
+/obj/item/storage/backpack/marine/satchel/scout_cloak/Dispose()
+	camo_off()
+	wearer = null
+	processing_objects.Remove(src)
+	. = ..()
```
Rohesie:
```suggestion
	return ..()
```

code/game/objects/items/storage/backpack.dm
```diff
@@ -420,14 +419,33 @@
...
+
+/obj/item/storage/backpack/marine/satchel/scout_cloak/dropped(mob/user)
+	camo_off(user)
+	wearer = null
+	processing_objects.Remove(src)
+	. = ..()
```
Rohesie:
```suggestion
	return ..()
```

code/game/objects/items/storage/backpack.dm
```diff
@@ -446,19 +464,20 @@
...
 		return
 
-	if (!camo_ready)
-		to_chat(M, "<span class='warning'>Your thermal dampeners are still recharging!")
+	if (camo_cooldown_timer)
+		to_chat(M, "<span class='warning'>Your thermal cloak is still recalibrating! It will be ready in [(camo_cooldown_timer - world.time) * 0.1] seconds.")
 		return
 
-	camo_ready = 0
 	camo_active = 1
```
Rohesie:
```suggestion
	camo_active = TRUE
```

code/game/objects/items/storage/backpack.dm
```diff
@@ -471,8 +490,8 @@
 	spawn(1)
 		anim(M.loc,M,'icons/mob/mob.dmi',,"cloak",,M.dir)
 
-	camo_active_timer = world.timeofday + SCOUT_CLOAK_TIMER
-	process_active_camo(usr)
+	processing_objects.Add(src)
+
 	return 1
```
Rohesie:
```suggestion
	return TRUE
```

code/game/objects/items/storage/backpack.dm
```diff
@@ -495,30 +514,60 @@
...
-			camo_off(user)
+	to_chat(user, "<span class='warning'>[details.Join(" ")]</span>")
 
-		sleep(10)	// Process every second.
+/obj/item/storage/backpack/marine/satchel/scout_cloak/item_action_slot_check(mob/user, slot)
+	if(!ishuman(user))
+		return FALSE
+	if(slot != WEAR_BACK)
+		return FALSE
+	return TRUE //only give action button when armor is worn.
```
Rohesie: Armor?

code/modules/projectiles/updated_projectiles/guns/specialist.dm
```diff
@@ -153,11 +153,17 @@
...
+						/obj/item/attachable/extended_barrel,
 						/obj/item/attachable/suppressor,
 						/obj/item/attachable/verticalgrip,
 						/obj/item/attachable/angledgrip,
 						/obj/item/attachable/bipod,
-						/obj/item/attachable/compensator)
+						/obj/item/attachable/compensator,
+						/obj/item/attachable/lasersight,
+						/obj/item/attachable/attached_gun/grenade,
+						/obj/item/attachable/attached_gun/shotgun,
```
Rohesie: Standard question: have to tested to see how those sprites look?

Rohesie:
The burst delay was added to compensate the removed fire delay. The logic would be that if you are not scoping while burst-firing, it would take a little time to reposition the gun and fire again. Of course it was poorly done, as the delay is applied between each shot, and not at the end of the burst.

Removing it is fine, but find another penalty that makes sense instead, to compensate.

code/modules/projectiles/updated_projectiles/guns/specialist.dm
RandomColonist: EB misaligned, can confirm.

Surrealaser: Fixed

code/__DEFINES/objects.dm
Rohesie: Planning on doing this?

code/__DEFINES/objects.dm
Rohesie: And this?

code/modules/mob/living/carbon/human/human_damage.dm
Rohesie: And this.

code/__DEFINES/objects.dm
Surrealaser: This is a style thing; I think it's fine.

code/__DEFINES/objects.dm
Surrealaser: Isn't it already headed?

Also what is meant by a 'different name'? I think SCOUT_CLOAK is pretty clear.

spookydonut: `SCOUT_CLOAK_OFF_DAMAGE` and `SCOUT_CLOAK_OFF_ATTACK` are bitflags whereas the others are actual values.   They should have a name that differentiates them as bitflags and be separated for the other things that arent bitflags

PR #658, opened 2018-11-12
Re-adding Gorg's walls/windows/frames after receiving permission to use them

Arachnidnexus:
New files are located at:
icons/turf/gorg_almayer.dmi
icons/turf/walls/gorg_prison.dmi
icons/obj/structures/gorg_windows.dmi
icons/obj/structures/gorg_window_frames.dmi

Thanks again to Gorg for letting the server use their sprites.


Arachnidnexus:
Removed the research containment walls, should be ready for merge unless the files themselves should be named something else/placed somewhere else.

PR #672, opened 2018-11-13
Prevents Warriors and Defenders from ventcrawling

DominikPanic:


DominikPanic:
> 
> 
> They still can ventcrawl through verb.

Good catch, should be fixed now.

Surrealaser:
Not a fan of the ban on Defender vent crawling and think it'll create problems with this caste where it spends the majority of playtime traversing the map, but we'll see how it goes.

DominikPanic:
> 
> 
> This has other effects besides ventcrawling. It affects staggerstun. We'll have to rethink that.

Fixed.

DominikPanic:
Welp turns out you can (this is what I get for not being a xeno main) anyway I figured out a shitty solution.

spookydonut:
#695 adds can_vent_crawl

code/modules/mob/living/carbon/xenomorph/xeno_helpers.dm
```diff
@@ -2,10 +2,14 @@
 
 
 /mob/living/carbon/Xenomorph/can_ventcrawl()
-	return (mob_size != MOB_SIZE_BIG)
+	if(mob_size == MOB_SIZE_BIG)
+		return FALSE
+	if(isXenoWarrior(src) || isXenoDefender(src))
```
Rohesie: Good idea. A more efficient way to do this would be by:
```
/mob/living/carbon/Xenomorph/Warrior/can_ventcrawl()
	return FALSE

/mob/living/carbon/Xenomorph/Defender/can_ventcrawl()
	return FALSE
```

PR #673, opened 2018-11-13
Removes armor clothing restrictions

DominikPanic:


Alterist:
marine_uniforms has suit_restricted vars you should take a look at too

DominikPanic:
> 
> 
> marine_uniforms has suit_restricted vars you should take a look at too

Fixed.

PR #678, opened 2018-11-14
Adds MP channel to command headsets. All channels start enabled.

DominikPanic:


Ghommie:
The channels starting off was done to prevent command personnel from being flooded by radio messages from sections they weren't overseeing. This applies mostly on highly populated servers, but still worth preserving.

DominikPanic:
Fixed @Rohesie

Rohesie:
Whoa, this wasn't necessary, but I love you having done it. ^^

PR #684, opened 2018-11-15
Explosive Pouch Added to Marine/SL/Spec Vendors

Surrealaser:
Per title.


carlarc:
im against putting it in the SL vendor since it kinda baits newbie sls to get an explosive pouch instead of a large general pouch which has the same storage space.
maybe make the pouch store 4 grenades? it seems kinda weird to me rn, just enough storage to be almost but not really worth it

Rohesie:
The problem is balancing on how many rockets in comparison with nades you'll store. Because rockets are much more valuable.

Surrealaser:
@Rohesie tbh I don't think it would be problematic to increase the capacity by +1

Rohesie:
Go for it then.

PR #687, opened 2018-11-15
Moves config to example files and gitignores the live used files.

spookydonut:


ninjanomnom:
We'll handle this the same as tg and we've never had any issues with the setup. The issue is just the current in between setup.

PR #688, opened 2018-11-15
Add files via upload

Hughgent:
This changes the pill machine buffer from 120 to 240.


Hughgent:
but it looks terrible because I am unfamiliar with Github

PR #693, opened 2018-11-16
Motion Detector Auto-Deactivates If User Dead

Surrealaser:
Per title.


StolichnayaElite:
@Shahhashun on Discord - While I see the merit in the argument for saving resources used by the MD, a fallen Motion Detector being creepy and pinging in the other room while you're shitting bricks down the hallway is a big part of the Aliens franchise. See the original AVP game that came out in 1999 for example.

I must respectfully disagree with this change.

code/game/objects/items/devices/motion_detector.dm
```diff
@@ -70,26 +71,43 @@
...
-		human_user = loc
-	else
-		active = FALSE
-		update_icon()
-		processing_objects.Remove(src)
+	if(!operator)
+		deactivate()
+		return
+
+	if(!loc == operator)
```
Rohesie:
```suggestion
	if(get_turf(src) != get_turf(operator))
```
Though all these checks could better go in a single `if` statement in between `||`

Surrealaser:
@Rohesie Looks like it, unless you can find a reason it isn't.

PR #700, opened 2018-11-17
Scout Tweak

Surrealaser:
1: Walk transparency increased to 85%

2: Fixed a minor bug with cloaking.

3: Scout gets a motion detector instead of plastique... so he can y'know... scout.


code/game/objects/items/storage/backpack.dm
```diff
@@ -496,6 +494,9 @@
 
 /obj/item/storage/backpack/marine/satchel/scout_cloak/proc/camo_off(var/mob/user)
 	if (!user)
+		camo_active = 0
```
Rohesie:
```suggestion
		camo_active = FALSE
```

RandomColonist:
Motion detector is actually very useful for a scout, but it's difficult to use with spec kit. Some suggestions about it and for scout in general:
[*] Helmet/armor integrated motion detector. Since cape requires both armor and helmet be equipped to work as a system - why not?
[*] Compact advanced version of motion detector that can work from pocket but has only short range detection (or without restrictment).
[*] Silent version of motion detector for stealth or in stealth mode (only for two suggestions above and specifically scout).
Suggestions above aimed for make scout actually not just stealth guy, but real Scout who can be aware of a treat. Also:
[*] In order, to make advanced smoke grenades more common i purpose to add them to scout spec kit/vendor, since spec also will have explosives pouch from vendor he will be able to carry it just fine, two-three enough for a kit.

Surrealaser:
@RandomColonist Scout now has access to a special silent recon version tactical sensor that fits into the pocket, as well as cloak grenades.

RandomColonist:
> @RandomColonist Scout now has access to a special silent recon version tactical sensor that fits into the pocket, as well as cloak grenades.

Yeah, that's great, but now i wounder how do i carry all that stuff and ammo without second pouch. Scout seems more restricted in carry capacity now, but that a question for other day.

PR #702, opened 2018-11-17
Ravager Now Has a Rage-O-Meter in its Stat Tab Also Rage GainZ From Damage Procs Limited

Surrealaser:
Per title.

Ravager can only gain rage from a damage instance once per 0.2 seconds.

Ravager base speed reduced to -0.4 down from -0.5


code/modules/mob/living/carbon/xenomorph/Castes/Ravager.dm
```diff
@@ -48,6 +48,12 @@
 		/datum/action/xeno_action/second_wind,
 		)
 
+/mob/living/carbon/Xenomorph/Ravager/Stat()
+	if (!..())
```
Rohesie: Use something like:
```
. = ..()
if(!.)
```
Also `TRUE` and `FALSE` here, for the returns.

code/__DEFINES/mobs.dm
```diff
@@ -486,7 +486,7 @@ var/list/global_mutations = list() // list of hidden mutation things
 #define RAV_CHARGEDISTANCE				7
 #define RAV_CHARGECOOLDOWN				300
 #define RAV_CHARGE_TYPE					3
-#define RAV_SECOND_WIND_COOLDOWN		600
+#define RAV_SECOND_WIND_COOLDOWN		3000
```
Rohesie:
```suggestion
#define RAV_SECOND_WIND_COOLDOWN		5 MINUTES
```

code/__DEFINES/mobs.dm
```diff
@@ -486,7 +486,7 @@ var/list/global_mutations = list() // list of hidden mutation things
 #define RAV_CHARGEDISTANCE				7
 #define RAV_CHARGECOOLDOWN				300
```
Rohesie:
```suggestion
#define RAV_CHARGECOOLDOWN				30 SECONDS
```

code/__DEFINES/mobs.dm
```diff
@@ -486,7 +486,7 @@ var/list/global_mutations = list() // list of hidden mutation things
...
 #define RAV_CHARGE_TYPE					3
-#define RAV_SECOND_WIND_COOLDOWN		600
+#define RAV_SECOND_WIND_COOLDOWN		3000
 #define RAV_RAVAGE_COOLDOWN				100
```
Rohesie:
```suggestion
#define RAV_RAVAGE_COOLDOWN				10 SECONDS
```

code/modules/mob/living/carbon/xenomorph/Castes/Ravager.dm
```diff
@@ -48,6 +49,13 @@
 		/datum/action/xeno_action/second_wind,
 		)
 
+/mob/living/carbon/Xenomorph/Ravager/Stat()
+	. = ..()
+	if(!.)
+		return FALSE
+
+	stat(null, "Rage: [rage] / [RAVAGER_MAX_RAGE]")
+	return TRUE
```
Rohesie: Is this needed? We are already preserving the return value in `.`
Same regarding the earlier `return FALSE`. If the `if` statement isn't true, then it will return null or FALSE anyway.

Rohesie:
I'd like xeno datums to be finished first, to avoid conflicts.

code/modules/mob/living/carbon/xenomorph/Castes/Ravager.dm
Rohesie: Uhh, I see I dun goofed by asking both preserving the return and to return TRUE or FALSE. My bad here.

PR #704, opened 2018-11-17
New Lobby Music

Hughgent:
Adds "Space Hero" from "Carmen Miranda's Ghost" to the lobby rotation music.

https://youtu.be/7XIdoUnhVWE

Cuts off after the lyric 
"Well the odds are good that you will be dead by twenty two, so you will be young and handsome all your life." 
Which is just before the last refrain.
"Hero is a four letter word."


Rohesie:
Probably should make a vote in #democracy about this.

Hughgent:
Please do so. Us normal folks can't write in #democracy.

PR #705, opened 2018-11-17
Edgy armor [Sprites only]

Arachnidnexus:
image
Not currently linked to anything, just added to the relevant files. Was originally intended to replace the Death Squad but _shrug_.


Rohesie:
Gloves?

PR #706, opened 2018-11-17
Tank Tweak (Runs over more objects)

Suicidalpickles:
allows the tank to run over:

-disposal units
-security camera consoles
-sec record consoles
-hydro trays


Surrealaser:
I like it, though I would have it add random debris as well, like metal rods, glass, robogibs and so on

Rohesie:
`Dispose()` is definitely not the right proc.

Suicidalpickles:
couldn't figure out how to get it running with a destroy proc in place of the dispose proc, kept coming up as undefined.
tldr im terribad

Rohesie:
You'll want something like:
```
new /obj/item/stack/sheet/metal(S.loc, 2)
cdel(S)
```
Though I'd wait for Stoned's MC port, and then do `qdel(S)` instead.

Rohesie:
Now that'd be ideal, but a bit beyond what Suicidalpickles can do.

PR #708, opened 2018-11-18
MedVac Bed/Beacon Improvements

Surrealaser:
-Can now link the medvac beacon to the medvac stretcher by hitting it with a folded bed.

-Medvac Beacon now gives information on whether it's linked and powered, as well as the coordinate and area data of the medvac bed it's linked to.


Rohesie:
Another folder to delete...

PR #718, opened 2018-11-19
New Sniper Ability: Targeting Laser

Surrealaser:
-Sniper gains the ability to use a laser aim mode while zoomed in with his sniper rifle.

-Toggle between laser aim and standard mode using Unique Action.

-While in laser mode, click a target to mark it with your laser; the target remains marked until you leave zoom or the target leaves your sight. You will be ready to fire immediately after marking a target.

-While you have a marked target, your shots will receive a significant boost in accuracy, ignore all other mobs and deal 50% bonus damage and pen to the marked target.

-Sniper also gets tactical binos instead of standard binos, and 4 cloak grenades.

-Guns now have their own zoom proc so that scopes can properly modify accuracy while zoomed.


code/game/objects/effects/overlays.dm
```diff
@@ -148,6 +148,17 @@
...
+/obj/effect/overlay/temp/sniper_laser
+	name = "laser"
+	mouse_opacity = 0
+	luminosity = 2
+	icon = 'icons/obj/items/projectiles.dmi'
+	icon_state = "sniper_laser"
+
+/obj/effect/overlay/temp/blinking_laser/Dispose()
+	SetLuminosity(0)
+	. = ..()
```
Rohesie:
```suggestion
	return ..()
```

code/modules/projectiles/updated_projectiles/gun_attachables.dm
```diff
@@ -1159,3 +1145,11 @@ obj/item/attachable/attack_hand(var/mob/user as mob)
 	return TRUE
 
 
+/obj/item/weapon/gun/zoom(mob/living/user, tileoffset = 11, viewsize = 12) //this is so the accuracy modifiers for the scopes apply correctly
+	. = ..()
+	if(istype(src.rail,/obj/item/attachable/scope))
```
Rohesie:
```suggestion
	if(istype(rail,/obj/item/attachable/scope))
```

code/modules/projectiles/updated_projectiles/gun_attachables.dm
```diff
@@ -1159,3 +1145,11 @@ obj/item/attachable/attack_hand(var/mob/user as mob)
...
+		var/obj/item/attachable/scope/m42a/S = src.rail
```
Rohesie:
```suggestion
		var/obj/item/attachable/scope/m42a/S = rail
```

code/modules/projectiles/updated_projectiles/guns/specialist.dm
```diff
@@ -40,6 +46,200 @@
 		update_attachables()
 		S.icon_state = initial(S.icon_state)
 
+/obj/item/weapon/gun/rifle/sniper/M42A/Fire(atom/target, mob/living/user, params, reflex = 0, dual_wield)
+	if(!able_to_fire(user))
+		return
+	if(targetlaser_on)
+		if(istype(target,/mob) )
```
Rohesie:
```suggestion
		if(ismob(target))
```

code/modules/projectiles/updated_projectiles/guns/specialist.dm
```diff
@@ -40,6 +46,200 @@
...
+			if(laser_target)
+				laser_target.vis_contents -= laser
+			cdel(laser)
+			to_chat(user, "<span class='danger'>You focus your targeting laser on [target]!</span>")
+			laser_target = target
+			var/obj/effect/overlay/temp/sniper_laser/LT = new (laser_target, "targeting laser")
+			laser = LT
+			targetlaser_on = FALSE
+			laser_target.vis_contents += laser
+			processing_objects.Remove(src) //So we don't accumulate additional processing.
```
Rohesie: Wouldn't this refer to the previous laser?

code/modules/projectiles/updated_projectiles/guns/specialist.dm
```diff
@@ -40,6 +46,200 @@
...
+
+/obj/item/weapon/gun/rifle/sniper/M42A/dropped()
+	laser_off()
+	. = ..()
+
+/obj/item/weapon/gun/rifle/sniper/M42A/process()
+	if(!zoom)
+		laser_off()
+		return
+	if(!istype(loc,/mob/living) )
```
Rohesie:
```suggestion
	if(!isliving(loc) )
```

code/modules/projectiles/updated_projectiles/guns/specialist.dm
```diff
@@ -40,6 +46,200 @@
...
+		laser = null
+		processing_objects.Remove(src)
+		return
+	if(!can_see(user, laser_target, length=23))
+		laser_target.vis_contents -= laser
+		laser_target = null
+		cdel(laser)
+		laser = null
+		processing_objects.Remove(src)
+		to_chat(world, "<span class='danger'>You lose sight of your target!</span>")
```
Rohesie:
```suggestion
		to_chat(user, "<span class='danger'>You lose sight of your target!</span>")
```

code/modules/projectiles/updated_projectiles/guns/specialist.dm
```diff
@@ -40,6 +46,200 @@
...
+		return
+
+/obj/item/weapon/gun/rifle/sniper/M42A/zoom(mob/living/user, tileoffset = 11, viewsize = 12) //tileoffset is client view offset in the direction the user is facing. viewsize is how far out this thing zooms. 7 is normal view
+	. = ..()
+	var/obj/item/attachable/scope/m42a/S = src.rail
```
Rohesie:
```suggestion
	var/obj/item/attachable/scope/m42a/S = rail
```

code/modules/projectiles/updated_projectiles/guns/specialist.dm
```diff
@@ -40,6 +46,200 @@
...
+	if(zoom)
+		S.accuracy_mod = config.max_hit_accuracy_mult
+	else
+		S.accuracy_mod = 0
+		if(user.client)
```
Rohesie:
```suggestion
```

code/modules/projectiles/updated_projectiles/guns/specialist.dm
```diff
@@ -40,6 +46,200 @@
...
+			user.client.change_view(viewsize)
```
Rohesie:
```suggestion
		user.client?.change_view(viewsize)
```

code/modules/projectiles/updated_projectiles/projectile.dm
```diff
@@ -521,6 +526,9 @@ Normal range for a defender's bullet resist should be something around 30-50. ~N
 		to_chat(world, "<span class='debuginfo'>Initial armor is: <b>[armor]</b></span>")
 		#endif
 		var/penetration = P.ammo.penetration > 0 || armor > 0 ? P.ammo.penetration : 0
+		if(src == P.shot_from.sniper_target(src))
```
Rohesie: This should be handled in the gun, not in defense.

code/modules/projectiles/updated_projectiles/projectile.dm
```diff
@@ -627,7 +635,12 @@ Normal range for a defender's bullet resist should be something around 30-50. ~N
 			#endif
 
 		var/penetration = P.ammo.penetration > 0 || armor > 0 ? P.ammo.penetration : 0
+		if(src == P.shot_from.sniper_target(src))
```
Rohesie: This should be handled in the gun, not in defense.

Rohesie (changes requested):

code/modules/projectiles/updated_projectiles/projectile.dm
Surrealaser: How would you make it do more damage from the gun-side specifically without having unintended effects... like say doing way more damage than you should to a wall (which is why I set it up like this)?

code/modules/projectiles/updated_projectiles/projectile.dm
Surrealaser: How would you make it do more damage from the gun-side specifically without having unintended effects... like say doing way more damage than you should to a wall (which is why I set it up like this)?

code/modules/projectiles/updated_projectiles/guns/specialist.dm
Surrealaser: No.

The previous laser is handled and removed by the code above.

code/modules/projectiles/updated_projectiles/projectile.dm
Rohesie: Adding a damage modifier to the gun while there is a targeted mob.

code/modules/projectiles/updated_projectiles/guns/specialist.dm
```diff
@@ -21,24 +21,223 @@
...
+		var/obj/item/attachable/sniperbarrel/Q = new(src)
+		Q.Attach(src)
+		update_attachables()
+		S.icon_state = initial(S.icon_state)
+
+/obj/item/weapon/gun/rifle/sniper/M42A/Fire(atom/target, mob/living/user, params, reflex = 0, dual_wield)
+	if(!able_to_fire(user))
+		return
+	if(targetlaser_on)
+		if(ismob(target))
```
Rohesie: Early return here.

code/modules/projectiles/updated_projectiles/guns/specialist.dm
Rohesie: This logic is a little muddy. Is the intention here to switch from one target to another?
Why is this removing one laser if one exists and creating a new one? If it is to refresh it in case the target is different, it should check that and only do it when it is.

Rohesie (changes requested):

code/modules/projectiles/updated_projectiles/projectile.dm
Surrealaser: That doesn't address the problem of doing more damage to walls, and other non-mob structures you can hit than you should.

Rohesie: Can you even target those? If so, why?

Rohesie (changes requested):

PR #724, opened 2018-11-20
Add a pen to Lower Chemistry and rearrange so the BS beaker is on top.

Hughgent:
as per title.

I hope I am not crazy as what I did was add the pen on top, then added a Bluespace beaker on top, then deleted the previous Bluespace beaker.

There seem to be more changes to the inner working code than should be. But this could just be a feature of Dream Maker.


Alterist:
Yeah, that happens.

This might be a bit presumptuous of me to ask, but I'm already working on edits to the Almayer map. Could you close this PR, and I'll add it to my edit?

Hughgent:
> 
> 
> Yeah, that happens.
> 
> This might be a bit presumptuous of me to ask, but I'm already working on edits to the Almayer map. Could you close this PR, and I'll add it to my edit instead?

Deal if you teach me the basics of mapping. I got an idea in me that I would like to see to fruition.

PR #725, opened 2018-11-20
New Weeds

Arachnidnexus:
Replacing weed sprites slowly and painfully. This PR covers weeds, nests, resin holes, purple sacs, and sticky resin.

image
image

_Edit:_ Only the first five base tiles are variants, the other 10 variants are just copy-pasted of those first 5. **Someone else needs to fix the weed randomizing code so it only uses the first five full tile variants randomly.**

_Edit 2:_ The icons in the UI need to be improved, purple sac could also be darkened a bit. Maybe use darkest purple from nest to outline purple sac base a little?


Arachnidnexus:
Closing because everything got moved to a new place.

PR #731, opened 2018-11-21
Re-allow the IMS as a necrosis surgery tool.

Hughgent:
Scalpel essentially becomes useless. But it's a small loss.


Hughgent:
ah dammit

PR #736, opened 2018-11-21
Changes Claymore Description 

Alterist:
Removes a bit of odd text for the commander's sword in the display case that will be in the almayer edit


Arachnidnexus:
Is the description now "When you were partying, I studied the blade. When you were having premarital sex, I mastered the blockchain. While you wasted your days at the gym in pursuit of vanity, I cultivated inner strength. And now that the world is on fire and the barbarians are at the gate you have the audacity to come to me for help." ?

Rohesie:
XP

PR #740, opened 2018-11-22
M3 Armor Variants and Marine Vendor Tweak

Rohesie:
* 6 armors for PFCs made using the sprite variants: regular, light, heavy, ballistic, melee and extended storage.
* Req vendor attachments re-ordered.
* Lots of new options added to the spare clothes vendor. Color of items tweaked.
* Shoulder holster added to the webbing options.
* Support roles have the brown webbing, combat ones the black one. Pure aesthetics.
* Essential gear spawning re-ordered, so you have boots and uniforms before less important items.
* Superfluous items that can be bought in the marine vendors, like syringes and signalers, removed from the essential gear.


Rohesie:
`/obj/machinery/vending/uniform_supply` is not affected by `cm_initialize.dm`
It will profit from a rework, but that will have to wait until we port TGUI

PR #745, opened 2018-11-23
SMG Ammo Cap Tweak

Surrealaser:
SMG ammo capacity increased to 60 rounds (extended to 90), so it actually has significantly more ammo than the M41A.


carlarc:
wacky but good

RandomColonist:
Lore died but it makes sense.

Hughgent:
image
if this is permanent, something to fix is the description.

PR #746, opened 2018-11-23
Lobby Screen DM;CA Removal

Rohesie:
* Stores lobby screens that feature DM;CA as a name. Kept for historical purposes.
* Randomly chooses from the screens that do not reference it directly.
* Kept two of the intermission screens as alternative icon states despite the name, but not on the random rotation.


Rohesie:
Case by case basis, I guess. We could start improving the quality of the lobby screens, really.

ActualRory:
I'd keep the one with the melted airlock.
It's cool looking, and a little subtle DMCA somewhere would be neat.

PR #750, opened 2018-11-24
Crusher Side Armor Nerf

Surrealaser:
-Crusher side armor reduced by 33% of its base armour.


Rohesie:
This induces a series of problems:
1) Moonwalking is more vital than ever. Is this a concept we want to keep?
2) The caste's armor no longer refers to anything. It used to be the side armor, while the front and back got minus/plus 50%, but right now it's just a reference value. This could be simplified,

PR #757, opened 2018-12-02
Temporary measure to get all logs.

DominikPanic:
Please do not look at this code. This code is shy and doesn't want to be seen.


Rohesie:
XP

DominikPanic:
> 
> 
> XP

Yeah, OK, Rohesie, I guess you could consider that a label. What I'd consider a label would be something acutally meaningful and engaging for the contributors. In reality, half of the people reading this sat out confused, baffled, unable to participate in this open-source model while the lucky maintainers just managed to get all the fun with the PRs. If this is going to be a pattern for your "reviews" (Maintainer Abuse) then I'd prefer you just stopped while you're ahead. Next time this happens, there will be an maintainer complaint. You have been warned.

